PR #30, opened 2020-01-31
Next not defined (fix #27)

### LINKED ISSUE #27: linting error: 'next' is not defined
## Describe the bug

Running the linting task returns the following error:

/Users/jo/Documents/git/wecount.inclusivedesign.ca/nuxt.config.js
  39:13  error  'next' is not defined  no-undef 

If `next` is a global function, this needs to be indicated as a global variable. However, it may just be something that actually hasn't been defined. If that's the case I wonder if it's related to https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/19

## To Reproduce

1. From the command line run `npm run lint`
2. The linting error for nuxt.config.js should be returned

## Expected behavior

Linting should return without errors.

TeddTech:
resolved issue #27


nuxt.config.js
```diff
@@ -35,9 +35,7 @@ export default {
 					})
 
 					callback(null, pageRoutes.concat(postRoutes))
-				}), function (err) {
-					return next(err)
-				})
+				}))
```
greatislander: @TeddTech I'm not sure this adequately addresses #27. By removing the error callback, this means that if the API is unavailable for any reason, the site will just break. Any thoughts on how you could handle this error?

TeddTech: I can import express for this functionality?

nuxt.config.js
```diff
@@ -35,8 +35,9 @@ export default {
 					})
 
 					callback(null, pageRoutes.concat(postRoutes))
-				}), function (err) {
-					return next(err)
+				}))
+				.catch((error) => {
+					throw new Error(error.response)
```
jobara: Based on the [Docs](https://nuxtjs.org/api/configuration-generate/#function-with-a-callback) it would seem that you should be passing the `callback` as the argument to `catch`.

TeddTech: Should I use `.catch(callback)` as a viable solution for now

PR #55, opened 2020-02-13
Main article

### LINKED ISSUE #46: Component: Main / Article
Article component.

TeddTech:
resolves #46

### DIFF
--- a/components/Article.vue
+++ b/components/Article.vue
@@ -1,0 +1,23 @@
+<template>
+	<div>
+		<h1 class="title">
+			<b>{{ Title }}</b>
+		</h1>
+		<p id="api-content" v-html="Content" />
+	</div>
+</template>
+
+<script>
+export default {
+	props: {
+		Title: {
+			type: String,
+			default: "Untitled Post"
+		},
+		Content: {
+			type: String,
+			default: "We Count content"
+		}
+	}
+}
+</script>
--- a/components/Article.vue
+++ b/components/Article.vue
@@ -1,10 +1,10 @@
 <template>
-	<div>
+	<article>
 		<h1 class="title">
-			<b>{{ Title }}</b>
+			{ Title }}
 		</h1>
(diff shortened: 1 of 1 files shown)


components/Article.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<div>
```
greatislander: For an article, can you suggest a more semantically appropriate container to use?

components/Article.vue
```diff
@@ -0,0 +1,23 @@
...
+		<h1 class="title">
+			<b>{{ Title }}</b>
```
greatislander: What is the purpose of the `<b>` tag here?

components/Article.vue
```diff
@@ -0,0 +1,23 @@
...
+		</h1>
+		<p id="api-content" v-html="Content" />
```
greatislander: Is this the best element to contain the content of a given page?

components/Article.vue
```diff
@@ -0,0 +1,23 @@
...
+<script>
+export default {
+	props: {
+		Title: {
+			type: String,
+			default: "Untitled Post"
+		},
+		Content: {
+			type: String,
+			default: "We Count content"
```
greatislander: What would be a better option for communicating that a page or post has no content?

components/Article.vue
```diff
@@ -0,0 +1,23 @@
...
+		<p id="api-content" v-html="Content" />
+	</div>
+</template>
+
+<script>
+export default {
+	props: {
+		Title: {
+			type: String,
+			default: "Untitled Post"
```
greatislander: This component will be used for posts and pages. Can we make this more generic?

jobara:
@TeddTech this [reference of HTML elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element) may be helpful.

PR #56, opened 2020-02-13
fix: moved main css to assets directory

### LINKED ISSUE #9: Assets: main CSS file should be stored in assets directory
## Describe the bug

As per this [document](https://nuxtjs.org/api/configuration-css), the source for the `main.css` file (currently located in `/css`) should be stored in the `/assets` directory.

TeddTech:
resolves #9


TeddTech:
> You need to update the Nuxt configuration for this to work.

Right! forgot about that

PR #68, opened 2020-02-15
feat: added NewsItem component

### LINKED ISSUE #63: Component: NewsItem
A component for rendering a post preview in the grid on the News and Views page.

### LINKED ISSUE #64: Page Layout: News
The News page, with a grid of `NewsItem` components and paged navigation.

### LINKED ISSUE #69: Component: NewsGrid
A grid of NewsItem components (#63) on the News page (#64).

TeddTech:
resolves #63
resolves #64
resolves #69


greatislander:
For reference:

components

pages/news-and-views/index.vue
```diff
@@ -5,34 +5,16 @@
...
-					<Post
-						v-else
-						:picture="x.picture"
-						:title="x.title"
-						:date="x.date"
-						:slug="x.slug"
-						style="width: 50%"
-					/>
-				</b-col>
+				<b-col id="api-content" v-for="x in row" :key="x.id" />
```
jobara: Is the `id="api-content"` supposed to be a `class`?

TeddTech: Yes, it is. Thanks for the catch!

TeddTech: The only reason why I edited this file in the PR was to delete the old `Post` component so it can pass the build.

greatislander: @TeddTech Can you consolidate the work from #70 and your unpushed branch for #64 into this PR? Since they all depend on each other it makes sense to use one pull request.

TeddTech: Yes, I agree. Working on it now

TeddTech:
> Please pull changes from #70 and the branch you've started for #64 into this PR so the changes can be evaluated as a whole.

This has been completed

TeddTech:
> I'm seeing errors when running `npm run dev`:
> 
> ```
> [Vue warn]: Invalid prop: type check failed for prop "date". Expected Date, got String with value "October 30, 2019".
> [Vue warn]: Invalid prop: type check failed for prop "picture". Expected Object, got String with value "http://wecount.inclusivedesign.ca/wp-content/uploads/2019/10/walrus_talks.png".
> [Vue warn]: Invalid prop: type check failed for prop "postList". Expected String with value "[object Object],[object Object]", got Array
> ```
> 
> Please address these issues.
> 
> Also, a few other notes:
> 
> 1. I'd like you to review the use of the `<article>` tag. The layout you're using here nests two levels of `<article>`, which it shouldn't. Why? Which level should be changed to a different tag, and what should the tag be?
> 2. When I run the [Validity](https://chrome.google.com/webstore/detail/validity/bbicmjjbohdfglopkidebfccilipgeif) Chrome Extension, I get an error related to your use of the `<div>` element within the `<figure>`, an error related to your use of the `<time>` element, and some errors related to `` tags. Please install the extension and see if you can find ways to address these. You'll need to read up about `<figure>` and `<time>`.
> 3. Please turn on [VoiceOver](https://help.apple.com/voiceover/mac/) on your Mac and see what's read out when you tab through the links on the News & Views page. Can this be improved further?

Hey Ned, 

I don't get those warning errors when I run `npm run dev`.

I looked into which correct tags I should use and `figure` seemed like the best fit for this use case. I also referenced multiple government of Canada pages and they use the same format. I would like to go over this tomorrow. In the meantime, I will address the issue with nested `article>` tags

components/NewsGrid.vue
```diff
@@ -0,0 +1,45 @@
+
+<template>
+	<div class="news-grid">
+		<div v-for="row in groupedList" :key="row.id" class="row">
+			<div v-for="x in row" :key="x.id" class="api-content">
```
jobara: Can we use more meaningful names instead of "x" and "row", something that would explain more clearly about what values they represent.

components/NewsGrid.vue
```diff
@@ -0,0 +1,45 @@
...
+				<NewsItem
+					v-if="row.length>1"
```
jobara: For readability, it's best to leave space around operators. Also see other comment about removing the condition.

components/NewsGrid.vue
```diff
@@ -0,0 +1,45 @@
...
+import _ from "lodash"
+import NewsItem from "~/components/NewsItem"
+export default {
+	components: {
+		NewsItem
+	},
+	props: {
+		postList: {
+			type: String,
+			default: () => []
```
jobara: Does this have to be a function, or can we just set the default to `[]`?

components/NewsItem.vue
```diff
@@ -0,0 +1,38 @@
...
+				<img class="date-icon">
+				<time>{{ date }}</time>
+			</div>
+		</figure>
+	</nuxt-link>
+</template>
+
+<script>
+export default {
+	// Pages have no media attached to them so when they render they are imageless
```
jobara: I'm a bit confused about this comment, or maybe what's expected. It says that the pages render imageless, but it looks like we add a default picture below.

components/NewsItem.vue
```diff
@@ -0,0 +1,38 @@
...
+	props: {
+		picture: {
+			type: Object,
+			default: () => "https://wecount.inclusivedesign.ca/wp-content/uploads/2019/10/We-Count-logos_colour-and-bw-01.png"
```
jobara: Similar to other comment. Does this have to be a function that returns a value? In the other props below, the default is just assigned the value directly.

components/NewsGrid.vue
jobara: I seem to have lost the comment about the conditional, so I'll add it here. It seems this conditional is setup so that an inline style can be added in some cases. I believe we want to avoid inline styles. I suggest for now to just remove the condition and the second `<NewsItem>`. We can explore ways to handle this with CSS when we get to the styling.

jobara:
> > I'm seeing errors when running `npm run dev`:
> > ```
> > [Vue warn]: Invalid prop: type check failed for prop "date". Expected Date, got String with value "October 30, 2019".
> > [Vue warn]: Invalid prop: type check failed for prop "picture". Expected Object, got String with value "http://wecount.inclusivedesign.ca/wp-content/uploads/2019/10/walrus_talks.png".
> > [Vue warn]: Invalid prop: type check failed for prop "postList". Expected String with value "[object Object],[object Object]", got Array
> > ```
> > 
> > 
> > Please address these issues.
> > Also, a few other notes:
> > 
> > 1. I'd like you to review the use of the `<article>` tag. The layout you're using here nests two levels of `<article>`, which it shouldn't. Why? Which level should be changed to a different tag, and what should the tag be?
> > 2. When I run the [Validity](https://chrome.google.com/webstore/detail/validity/bbicmjjbohdfglopkidebfccilipgeif) Chrome Extension, I get an error related to your use of the `<div>` element within the `<figure>`, an error related to your use of the `<time>` element, and some errors related to `` tags. Please install the extension and see if you can find ways to address these. You'll need to read up about `<figure>` and `<time>`.
> > 3. Please turn on [VoiceOver](https://help.apple.com/voiceover/mac/) on your Mac and see what's read out when you tab through the links on the News & Views page. Can this be improved further?
> 
> Hey Ned,
> 
> I don't get those warning errors when I run `npm run dev`.
> 
> I looked into which correct tags I should use and `figure` seemed like the best fit for this use case. I also referenced multiple government of Canada pages and they use the same format. I would like to go over this tomorrow. In the meantime, I will address the issue with nested `article>` tags

@TeddTech these errors are showing up for me when I visit the news-and-views page. 

Regarding the `<figure>` It looks like the issue is where you've placed the `<figcaption>` element. If you read the [spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-figure-element.html) it says:

> Content model:
    Either: One figcaption element followed by flow content.
    Or: Flow content followed by one figcaption element.
    Or: Flow content.

As @greatislander mentioned, you should run the pages through a validator.

components/NewsGrid.vue
TeddTech: how's `post` and `postRow`?

components/NewsGrid.vue
TeddTech: It has to be a function or else it throws a warning

components/NewsItem.vue
TeddTech: This should be removed now. Was added before default images were implemented.

components/NewsGrid.vue
greatislander: @TeddTech can you find an explanation of this in the documentation? That would be helpful.

components/NewsItem.vue
TeddTech: It has to be a function or else it throws a warning

components/NewsGrid.vue
TeddTech: This change has been implemented

components/NewsGrid.vue
TeddTech: These changes have been implemented

components/NewsGrid.vue
jobara: It just occurred to me that this is another artifact of trying to achieve the styling. Please correct me if I'm wrong. It looks like you are chunking the posts into a 2x2 array, then iterating over that to output a structure like:

```html
<div>
    <div>Post</div>
    <div>Post</div>
</div>
<div>
    <div>Post</div>
    <div>Post</div>
</div>
```

Which you are essentially using to create a two column layout of the News Items.

However, there may be a way to do this with styling, for example using [flex-box](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), [grid](https://css-tricks.com/snippets/css/complete-guide-grid/) or something else. Because we aren't worried about the styling, at the moment, let's simplify this by just outputting the posts and not worry about the chunking and columns for now.

components/NewsGrid.vue
TeddTech: > For readability, it's best to leave space around operators. Also see other comment about removing the condition.

After adding spaces around operators I get the following error:

`3:12  warning  Unexpected spaces found around equal signs  vue/no-spaces-around-equal-signs-in-attribute`

greatislander: Ref: https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/no-spaces-around-equal-signs-in-attribute.md

components/NewsGrid.vue
TeddTech: This is the error that is thrown when the default value is not in function form:

```  27:4  error  Type of the default value for 'postList' prop must be a function  vue/require-valid-default-prop```

File: components/NewsGrid.vue (hunk 3 above)
Comment (jobara) [reply components/NewsGrid.vue]: I should clarify, when I talk about operators I specifically meant `row.length>1` where the ">" has no whitespace. I didn't mean the "=" for the attribute assignments.

File: components/NewsGrid.vue (hunk 4 above)
Comment (jobara) [reply components/NewsGrid.vue]: Actually this is interesting, I think the problem here, why you are getting the error you just mentioned and why are seeing those errors on the news-and-views page, is that you set the type to `String`. Here the default is actually an array. It probably doesn't complain about the function because it is not evaluating the function when the linting happens. However, when it runs we get the error because the array is being returned.

Comment (TeddTech) [reply components/NewsGrid.vue]: @greatislander here is an explanation I found: 

> There will be a bug. because every instance will share the same thing when not passing the target property to this component.

https://github.com/vuejs/vue/issues/1032

File: components/NewsGrid.vue (hunk 2 above)
Comment (TeddTech) [reply components/NewsGrid.vue]: Okay, sounds good. I'll get started on that

**Comment by TeddTech** [comment]
> > > I'm seeing errors when running `npm run dev`:
> > > ```
> > > [Vue warn]: Invalid prop: type check failed for prop "date". Expected Date, got String with value "October 30, 2019".
> > > [Vue warn]: Invalid prop: type check failed for prop "picture". Expected Object, got String with value "http://wecount.inclusivedesign.ca/wp-content/uploads/2019/10/walrus_talks.png".
> > > [Vue warn]: Invalid prop: type check failed for prop "postList". Expected String with value "[object Object],[object Object]", got Array
> > > ```
> > > 
> > > 
> > > Please address these issues.
> > > Also, a few other notes:
> > > 
> > > 1. I'd like you to review the use of the `<article>` tag. The layout you're using here nests two levels of `<article>`, which it shouldn't. Why? Which level should be changed to a different tag, and what should the tag be?
> > > 2. When I run the [Validity](https://chrome.google.com/webstore/detail/validity/bbicmjjbohdfglopkidebfccilipgeif) Chrome Extension, I get an error related to your use of the `<div>` element within the `<figure>`, an error related to your use of the `<time>` element, and some errors related to `` tags. Please install the extension and see if you can find ways to address these. You'll need to read up about `<figure>` and `<time>`.
> > > 3. Please turn on [VoiceOver](https://help.apple.com/voiceover/mac/) on your Mac and see what's read out when you tab through the links on the News & Views page. Can this be improved further?
> > 
> > 
> > Hey Ned,
> > I don't get those warning errors when I run `npm run dev`.
> > I looked into which correct tags I should use and `figure` seemed like the best fit for this use case. I also referenced multiple government of Canada pages and they use the same format. I would like to go over this tomorrow. In the meantime, I will address the issue with nested `article>` tags
> 
> @TeddTech these errors are showing up for me when I visit the news-and-views page.
> 
> Regarding the `<figure>` It looks like the issue is where you've placed the `<figcaption>` element. If you read the [spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-figure-element.html) it says:
> 
> > Content model:
> > Either: One figcaption element followed by flow content.
> > Or: Flow content followed by one figcaption element.
> > Or: Flow content.
> 
> As @greatislander mentioned, you should run the pages through a validator.

I have refactored the figure tag in NewsItem component by taking the div tag outside it and placing it below

File: components/NewsItem.vue (hunk 6 above)
Comment (jobara) [reply components/NewsItem.vue]: Does this still have to be a function? Looks like it should be able to be a String.

File: components/NewsGrid.vue (hunk 4 above)
Comment (jobara) [reply components/NewsGrid.vue]: Could you please leave a comment in the code about this, with a link the to the issue you reference.

File: components/NewsGrid.vue [hunk 7]
```diff
@@ -0,0 +1,34 @@
...
+	props: {
+		postList: {
+			type: Array,
+			default: () => []
+		}
+	},
+	computed: {
+		// groupedList () {
+		// 	return _.chunk(this.postList, 2)
+		// }
```
jobara: Please remove the commented out code

components/NewsGrid.vue
greatislander: @TeddTech Ideally reference the Vue docs? There should be something in there about this.

components/NewsItem.vue
TeddTech: Yes, you're correct. I have changed it

components/NewsGrid.vue
TeddTech: All done

components/NewsItem.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+	<!-- This compenent is used by data from pages api call as well as post api call -->
+	<nuxt-link :to="'/news-and-views/' + slug" class="news-item">
+		<figure class="preview-media">
+			<img :src="picture" class="news-item-img" alt="">
+			<figcaption>
+				<h2>
```
greatislander: Move title out of the `<figure>` and remove the `<figcaption>`.

components/NewsItem.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+	<!-- This compenent is used by data from pages api call as well as post api call -->
+	<nuxt-link :to="'/news-and-views/' + slug" class="news-item">
```
greatislander: Make the containing element an `<article>`, put the link inside the `<h2>` surrounding the title.

components/NewsItem.vue
```diff
@@ -0,0 +1,40 @@
...
+		<figure class="preview-media">
+			<img :src="picture" class="news-item-img" alt="">
+			<figcaption>
+				<h2>
+					{{ title }}
+				</h2>
+			</figcaption>
+		</figure>
+		<div class="date">
+			<img class="date-icon" alt="">
```
greatislander: Remove this.

components/NewsItem.vue
```diff
@@ -0,0 +1,40 @@
...
+			<time>{{ date }}</time>
```
greatislander: Add the proper `datetime` attribute from the API data: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

components/NewsItem.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+	<!-- This compenent is used by data from pages api call as well as post api call -->
+	<nuxt-link :to="'/news-and-views/' + slug" class="news-item">
+		<figure class="preview-media">
+			<img :src="picture" class="news-item-img" alt="">
```
greatislander: Pull `alt` tag from the API.

components/NewsGrid.vue
greatislander: https://vuejs.org/v2/guide/components-props.html#Prop-Validation

TeddTech:
> Hi @TeddTech, this looks great. One small change: I think it would be better to name the prop for the alt tag `altTag` or something like that for clarity.

I have completed this change

TeddTech:
> I noticed one other thing. Each `<NewsItem>` is wrapped in a `<div>` within the grid container. Any reason why you can’t use `v:for` directly on the `<NewsItem>` component?

That's what I originally attempted to do. It throws an error saying you are not allowed to use vue directives in the directed child tag of a template tag. Also, template tags can only have a single child tag. I will recreate the error and post it below.

TeddTech:
> I noticed one other thing. Each `<NewsItem>` is wrapped in a `<div>` within the grid container. Any reason why you can’t use `v:for` directly on the `<NewsItem>` component?

This is the error it gives if you do that
```
[vue/valid-template-root]
The template root disallows 'v-for' directives.
```

PR #72, opened 2020-02-19
feat: simplified error-page

### LINKED ISSUE #71: Layout: error page
Improve error page to use simplified markup (Article component).

TeddTech:
using Article component

resolves #71

### DIFF
--- a/layouts/error.vue
+++ b/layouts/error.vue
@@ -1,28 +1,5 @@
 <template>
-	<div class="container">
-		<div v-if="error.statusCode === 404">
-			<h1>
-				Page not found
-			</h1>
-			<p>
-				The page you were looking for could not be found. It might have
-				<br>
-				been removed, renamed, or did not exist in the first place.
-			</p>
-		</div>
-		<div v-else>
-			<h1>
-				An error occurred
-			</h1>
-		</div>
-		<div>
-			<p>
-				<nuxt-link to="/">
-					Home page
-				</nuxt-link>
-			</p>
-		</div>
-	</div>
+	<Article :title="title" :content="content" />
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


layouts/error.vue
```diff
@@ -32,6 +9,12 @@ export default {
 			type: Object,
 			default: () => {}
 		}
+	},
+	data () {
+		return {
+			title: (this.error.statusCode === 404) ? "Page not found" : "An error occurred",
+			content: (this.error.statusCode === 404) ? "<p>The page you were looking for could not be found. It might have<br>been removed, renamed, or did not exist in the first place.</p><div><p><nuxt-link to='/'>Home page</nuxt-link></p></div>" : "<div><p><nuxt-link to='/'>Home page</nuxt-link></p></div>"
```
greatislander: A few questions:

- Why is there a `<br>` in the middle of the sentence?
- Why is there a `<div>` around some of the paragraphs?

TeddTech: * It is a long sentence and felt it deserved a line break
* There are divs to serparate `<div><p><nuxt-link to='/'>Home page</nuxt-link></p></div>` (the linke leading back to the home page) from the rest of the content

greatislander: @TeddTech:

>     * It is a long sentence and felt it deserved a line break

On different sized screens, this won't work. For example on a small phone, it'll look like this:

Screen Shot 2020-02-19 at 11 14 39 AM

It's better to use CSS to set the container width appropriately rather than forcing line breaks. Please remove the `<br />` tag.

>     * There are divs to serparate `<div><p><nuxt-link to='/'>Home page</nuxt-link></p></div>` (the linke leading back to the home page) from the rest of the content

Take a look at this: https://codepen.io/greatislander/full/XWbjXgM

There's no need for the extra `<div>`; `<p>` tags are block elements and provide the necessary separation.

TeddTech:
> I'm curious as to why you changed the `<nuxt-link>` components to `<a>` tags?

nuxt-link was not rendering as a link with the v-html directive in the `Article` component

layouts/error.vue
TeddTech: Okay, these changes have now been applied.

PR #73, opened 2020-02-20
feat: added pagination component

### LINKED ISSUE #65: Component: Pagination
Paged navigation for the News page.

TeddTech:

## Description

Add pagination component. This component should be used on individual posts and pages which list results, such as "News and Views" and search/tag pages.

## Criteria

- [x] Previous and next links must link to the previous/next post or page of results.
- [x] Previous and next link text must be consistent.
- [x] Previous link must only appear if there is one or more prior post or page of results.
- [x] Next link must only appear if there is one or more subsequent post or page of results.
- [x] If the current item is the first item in a sequence, the `current` link should be the first numeric page link.
- [x] If the current item is the last item in a sequence, the `current` link should be the last numeric page link.
- [x] Numeric page links must include `aria-label` attributes with the correct page number.
- [x] Ellipses should only appear in numeric page links if there are more than three pages of results before or after to the current item.
- [x] On "News and Views", search, and tag pages, the NewsGrid should be broken into chunks of 6 items and paginated accordingly.

## Steps to test

1. Visit a single news item.
2. Visit "News and Views" page.
3. Visit search results page for `inclusion`.
4. Visit tag page for `AI`.

**Expected behavior:** Paginated navigation should work according to the criteria above in all four contexts.

## Additional information

Not applicable.

## Related issues

Resolves #65.


components/PostArticle.vue
```diff
@@ -10,11 +10,17 @@
 				{{ t }}
 			</nuxt-link>
 		</div>
+		<hr class="lineBreak">
```
greatislander: Remove this line break.

components/Pagination.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<nav class="pagination is-centered" role="navigation" aria-label="pagination">
-		<a :href="beforeLink" class="pagination-previous">Previous</a>
-		<a :href="afterLink" class="pagination-next">Next</a>
+		<a v-if="currentPageNum>1" :href="beforeLink" class="pagination-previous">Previous</a>
```
greatislander: It's best practice to add spaces around the `<` / `>` operators.

TeddTech:
With the news-and-views page and tag page, the generation of the static pages will be straightforward because the amount of the results can be determined prematurely but not with the search page
the amount of the results depends on the `searchQuery` entered

 In terms of maintainability, a lot more upkeep of the code will have to be done using the `/page/1` format.

Every time there is a new tag the pages will have to be generated upon build `tag?s=AI&page=10`

greatislander:
> Every time there is a new tag the pages will have to be generated upon build 

This is also true every time there's a new news article, isn't it?

TeddTech:
> > Every time there is a new tag the pages will have to be generated upon build
> 
> This is also true every time there's a new news article, isn't it?

Every time there is a new post `page/1` will either go up to `page/2` or stay the same.
Every time there is a new tag, the routes for each `page/#` needs to be generated for each tag

This is impossible for search because of the infinite things a user can search

with the `/?page=1` there's no need to generate web pages during build

greatislander:
```
/news-and-views/index.vue # Page 1
/news-and-views/page/_num/index.vue # Page 2, 3, 4…
```

jobara:
> > > Every time there is a new tag the pages will have to be generated upon build
> > 
> > 
> > This is also true every time there's a new news article, isn't it?
> 
> Every time there is a new post `page/1` will either go up to `page/2` or stay the same.
> Every time there is a new tag, the routes for each `page/#` needs to be generated for each tag
> 
> This is impossible for search because of the infinite things a user can search
> 
> with the `/?page=1` there's no need to generate web pages during build

@greatislander and @TeddTech I think I've lost the context for this conversation and also not sure if there has been a resolution. Could you pease summarize and possibly add a code reference or snippet.

greatislander:
@jobara When the site is built statically, the news page can be chunked into pages of 10 posts per page because it's a known quantity (retrieved from the API). This means that Nuxt can be set up to generate nice clean URLs:

    /news-and-views/page/2

However, there's two other places where content needs to be paginated at present:

1. Search results. These are displayed at:

     /search?s=ai

They can be paginated, but that has to happen on the client side and use query params:

     /search?s=ai&page=2

2. Tag archives — pages which display all posts with a given tag. Right now these are displayed at:

     /tag?s=inclusion

The discussion with @TeddTech lead to the conclusion that rather than having tag archives created dynamically based on a query string, the WordPress API's [tags endpoint](https://wecount.inclusivedesign.ca/wp-json/wp/v2/tags) should be used to create these pages at build time, in this format:

     /tag/inclusion

That will also let them be paginated using nice clean URLs, because you can also use the WordPress API to get a [listing of posts with a given tag ID](https://wecount.inclusivedesign.ca/wp-json/wp/v2/posts?tags=4):

     /tag/inclusion/page/2

Let me know if this makes sense. @TeddTech will tackle the tag API changes in a separate PR as it's out of scope for this one.

greatislander:
@jobara When the site is built statically, the news page can be chunked into pages of 10 posts per page because it's a known quantity (retrieved from the API). This means that Nuxt can be set up to generate nice clean URLs:

    /news-and-views/page/2

However, there's two other places where content needs to be paginated at present:

1. Search results. These are displayed at:

     /search?s=ai

They can be paginated, but that has to happen on the client side and use query params:

     /search?s=ai&page=2

2. Tag archives — pages which display all posts with a given tag. Right now these are displayed at:

     /tag?s=inclusion

The discussion with @TeddTech lead to the conclusion that rather than having tag archives created dynamically based on a query string, the WordPress API's [tags endpoint](https://wecount.inclusivedesign.ca/wp-json/wp/v2/tags) should be used to create these pages at build time, in this format:

     /tag/inclusion

That will also let them be paginated using nice clean URLs, because you can also use the WordPress API to get a [listing of posts with a given tag ID](https://wecount.inclusivedesign.ca/wp-json/wp/v2/posts?tags=4):

     /tag/inclusion/page/2

Let me know if this makes sense. @TeddTech will tackle the tag API changes in a separate PR as it's out of scope for this one.

jobara:
@greatislander thanks for the explanation, that really helped. I'm guessing we don't have designs for this, but does the Wordpress API support sorting by multiple tags? That is for a posts that are tagged as "ai" and "inclusion" for example.

greatislander:
@jobara Yes, it does, but as you say I think we need to hold off on this until further design work has been completed (not to mention until we have more content).

TeddTech:
> ```diff
> wpDomain: "http://b0907dd6.ngrok.io",
> ```

It throws an error when implementing this change 

`Cannot read property '0' of undefined`

```
			let altTag = ""

			if (postType === "posts") {
				count += 1
				tags = r.data[i].pure_taxonomies.tags.map(({ name }) => name)
				const api = r.data[i]._links["wp:featuredmedia"][0].href // this line is throwing the error
				const pic = await axios.get(api)
				picture = pic.data.source_url
				altTag = pic.data.alt_text
			}
```

greatislander:
@TeddTech What does that error tell you?

TeddTech:
> @TeddTech What does that error tell you?

There is no data at `._links["wp:featuredmedia"]` for for the wpDomain: `"http://b0907dd6.ngrok.io"` api call

TeddTech:
Deploy preview is failing but everything works fine when I test it locally

```
Entrypoint app = server.js
ℹ Generating pages                                                                                                                                                                       14:11:02
✔ Generated /inclusion-challenges                                                                                                                                                        14:11:02
✔ Generated /tools                                                                                                                                                                       14:11:02
✔ Generated /our-data                                                                                                                                                                    14:11:02
✔ Generated /about                                                                                                                                                                       14:11:02
✔ Generated /                                                                                                                                                                            14:11:02
✔ Generated /news-and-views                                                                                                                                                              14:11:06
✔ Generated /news-and-views/walrus-talk                                                                                                                                                  14:11:06
✔ Generated /news-and-views/page/1                                                                                                                                                       14:11:06
✔ Generated /news-and-views/the-importance-of-peripheral-vision  
```

nuxt.config.js
```diff
@@ -25,7 +25,18 @@ export default {
 						}
 					})
 
-					callback(null, pageRoutes.concat(postRoutes))
+					const totalPages = posts.headers["x-wp-totalpages"]
```
greatislander: You're only getting a page count here, @TeddTech — you aren't fetching all the pages of posts. On line 11, `axios.get(`${Config.wpDomain}${Config.apiBase}posts`)` will just fetch the first page and then stop. I expect this might be part of the issue.

greatislander: (I know there's only one page of posts to fetch right now, but it's going to break for sure when there are more.)

TeddTech: I believe it's fixed now

greatislander: @TeddTech What happens if we have more than 100 posts?

TeddTech:
> Hi @TeddTech, this build is still failing on Netlify. We need to resolve this. Can you please clean up your latest commit, removing commented out code and adding documentation where appropriate?

Okay, I will do. This is strange it works locally for me.

TeddTech:
After inspecting the failing check it seems it is not passing due to a 500 error.

nuxt.config.js
```diff
@@ -5,27 +5,37 @@ export default {
...
-						return {
-							route: (page.slug !== "home") ? "/" + page.slug : "/",
-							payload: page
-						}
-					})
-
+				.then(axios.spread((posts) => {
+					// Add each post to the postRoutes variable
+					// replace line below with this when totalPages > 1:
+					// const postRoutes = posts.data.map((...post) => {
```
greatislander: This needs to be done for this pull request to be complete.

nuxt.config.js
```diff
@@ -5,27 +5,37 @@ export default {
...
 					const postRoutes = posts.data.map((post) => {
 						return {
 							route: "/news-and-views/" + post.slug,
 							payload: post
 						}
 					})
 
-					callback(null, pageRoutes.concat(postRoutes))
+					// What happened to the page routes (for /about, /tools, etc.)?
```
greatislander: I noticed that your latest changes to the Nuxt config remove the code which sets up the routes for pages. Is there a reason for this?

nuxt.config.js
TeddTech: What needs to be done?

nuxt.config.js
TeddTech: Yeah, the generate field should be used to build the dynamic pages only (pages where the parent folder name starts with underscore only, e.g. _webpage). All other regular pages are built automatically without specifying them in the generate field of nuxt.config.

nuxt.config.js
TeddTech: if you change `posts` to `...post` and `totalPages` <= 1 the code breaks

nuxt.config.js
greatislander: I think we should be generating the regular pages dynamically as well, but that's out of the scope of this pull request.

nuxt.config.js
greatislander: Can you write code that handles both situations? It will need to work no matter how many pages of results there are.

nuxt.config.js
TeddTech: They still are generated without specifying them in the generate field. You can test this out by running `npm run generate`. You can see the regular pages are still generated statically without specifying them

greatislander: How does that happen?

TeddTech: `npm run generate` automatically builds all pages statically excluding the dynamic pages. Those have to be specified in the generate field of nuxt.config

greatislander: How does it automatically build all pages statically, is my question?

PR #82, opened 2020-02-25
feat: add proper action route to search form

greatislander:

## Description

Sets the `action` attribute of the search form and the name of the search input to the correct values so that searching works.

## Steps to test

1. Check out this branch.
2. Run `npm install && npm run dev`.
3. Search for "inclusion".

**Expected behavior:** Search results are displayed.

## Additional information

Not applicable.

## Related issues

Not applicable.

### DIFF
--- a/components/SearchForm.vue
+++ b/components/SearchForm.vue
@@ -1,8 +1,8 @@
 <template>
 	<div class="search-container">
-		<form id="search-form" action="">
+		<form id="search-form" action="/search">
 			<label for="search">Search for:</label>
-			<input id="search" type="search" placeholder="Search..." name="search">
+			<input id="search" type="search" placeholder="Search..." name="s">
 			<button type="submit">
 				Search
 			</button>


components/SearchForm.vue
```diff
@@ -1,8 +1,8 @@
 <template>
 	<div class="search-container">
-		<form id="search-form" action="">
+		<form id="search-form" action="/search">
```
TeddTech: is `/search` a method that will have the proper search functionality?

components/SearchForm.vue
```diff
@@ -1,8 +1,8 @@
...
 			<label for="search">Search for:</label>
-			<input id="search" type="search" placeholder="Search..." name="search">
+			<input id="search" type="search" placeholder="Search..." name="s">
```
TeddTech: why `s` instead of `search`?

components/SearchForm.vue
greatislander: The search path, as defined in pages, is `search`: https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/master/pages/search/_query/index.vue

components/SearchForm.vue
greatislander: Because the search page uses the `s` URL parameter to determine the search query: https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/master/pages/search/_query/index.vue#L23

greatislander:
@TeddTech The search form, when submitted, makes an [HTTP request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) using the `GET` method to the URL specified as the form's `action` (see details [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form)). The fields in the form become URL parameters — in this case, `?s=searchterm`. [Here's some more info](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL) on how URLs are constructed and the various pieces.

PR #95, opened 2020-03-11
Styles links button (#95) (resolves #80)

### LINKED ISSUE #80: Create link and button styles

Implement styles for links and buttons as specified in Figma.

Using the appropriate Sass partials, implement link and button styles with hover and focus states.

Not applicable.

## Additional context

- https://css-tricks.com/a-complete-guide-to-links-and-buttons/

TeddTech:

## Steps to test

1. Look over button styles
2. Look over link styles

## Additional information

not applicable

## Related issues

resolves #80


assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+
+// Links and Buttons
+a {
+	color: #6895c6;
+	font-weight: bold;
+	text-decoration: none;
+}
+a:hover,
+a:focus {
+	background-color: #FADC90;
```
greatislander: This should be the appropriate color variable.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	font-weight: bold ;
+}
+p {
+	font-size: 1em ;
+	font-weight: regular ;
+}
+
+// Links and Buttons
+a {
+	color: #6895c6;
```
greatislander: This should be the appropriate color variable.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	font-weight: bold;
+	text-decoration: none;
+}
+a:hover,
+a:focus {
+	background-color: #FADC90;
+}
+
+button {
+	background-color: #cee6c5;
```
greatislander: This should be the appropriate color variable.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	border: 1px solid #8CAABB;
```
greatislander: This should be the appropriate color variable.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	border-radius: 70px;
+	box-sizing: border-box;
```
greatislander: You should add this as a global rule. See: https://css-tricks.com/box-sizing/

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	color: #6895c6;
+	font-weight: bold;
+	text-decoration: none;
+}
+a:hover,
+a:focus {
+	background-color: #FADC90;
+}
+
+button {
```
greatislander: The button does not appear as designed.

Figma:

Screen Shot 2020-03-10 at 6 47 36 PM

Chrome: 

Screen Shot 2020-03-10 at 6 48 23 PM

Safari:

Screen Shot 2020-03-10 at 6 48 37 PM

Firefox:

Screen Shot 2020-03-10 at 6 48 42 PM

Needs styling for font size, weight, and line height, padding, and height.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	background-color: #cee6c5;
+	border: 1px solid #8CAABB;
+	border-radius: 70px;
+	box-sizing: border-box;
+	transform: perspective(1px) translateZ(0);
+  }
+button:before {
```
greatislander: You should use a `box-shadow` for this instead of pseudo-content.

assets/css/base/_base.scss
```diff
@@ -1 +1,52 @@
...
+	border: 1px solid #8CAABB;
+	border-radius: 70px;
+	content: '';
+	opacity: 0;
+	position: absolute;
+  }
+button:hover:before, button:focus:before, button:active:before {
```
greatislander: You should use a `box-shadow` for this instead of pseudo-content.

greatislander:
@TeddTech I've added some stricter stylelint rules. You'll need to pull my latest commits and run `npm ci` to install them.

assets/css/base/_base.scss
TeddTech: Like this?

```css
* {
  box-sizing: border-box;
}
```

assets/css/base/_base.scss
TeddTech: I have tried. I am unable to duplicate the button styling of the Figma file with box-shadow. It allows for opaque shadows but there is no option to keep the border of the shadow present.

greatislander: Okay cool, pseudo-content works then.

TeddTech:
> @TeddTech Can you merge in the latest changes from the master branch and then we can take a look at this again? Thanks!

will do

assets/css/base/_base.scss
greatislander: Yep!

TeddTech:
Ready for review

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
 p {
   font-size: 1em;
 }
+
+// Links and Buttons
+a {
+  color: $blue-dark;
```
greatislander: @TeddTech I think this is the wrong colour — looking at Figma I believe the links are actually `#5e82ae` (`$blue-alt`).

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+  font-weight: bold;
```
greatislander: As was the case ion the typography pull request, this should be a numeric value. It might be a good idea to add variables for font weights to the variables file, so for example you could use `$bold` instead of having to remember that bold is 700. Does that make sense @TeddTech?

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+}
+
+button {
+  background-color: $green-light;
+  border: 2px solid $blue-alt-light;
+  border-radius: 70px;
+  font-size: 24px;
+  font-weight: 400;
+  height: 53px;
+  line-height: 34px;
```
greatislander: It's a good idea to use a unitless number for a line height so that the line height stays proportional if the user resizes the browser font size. See here: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values

The way you figure out this number is by dividing the line height in pixels by the font size in pixels, so in this case: 34 / 24. Sass can do this for you if you set the line height value as follows:

```suggestion
  line-height: (34 / 24);
```

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+a:hover,
+a:focus {
+  background-color: $orange-focus;
+}
+
+button {
+  background-color: $green-light;
+  border: 2px solid $blue-alt-light;
+  border-radius: 70px;
+  font-size: 24px;
```
greatislander: It's a great idea to use a [relative unit](https://hankchizljaw.com/wrote/relative-sizing-with-em-units/) for the font size here. If you look at Figma you can see that the `<h3>` style is `24px` which is equal to `1.5em`. You can figure out a font size in `em` by dividing the font size by the base font size (which is 16):

24px / 16px = 1.5em

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+
+a:hover,
+a:focus {
+  background-color: $orange-focus;
+}
+
+button {
+  background-color: $green-light;
+  border: 2px solid $blue-alt-light;
+  border-radius: 70px;
```
greatislander: It's a good idea to use `ems` here (see following comment). This makes sure the element will scale. You can use the formula (70px / 16px) &times; 1em to figure this out (here that works out to 4.375em).

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+}
+
+a:hover,
+a:focus {
+  background-color: $orange-focus;
+}
+
+button {
+  background-color: $green-light;
+  border: 2px solid $blue-alt-light;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+  border-radius: 70px;
+  font-size: 24px;
+  font-weight: 400;
+  height: 53px;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+  line-height: 34px;
+  padding: 9px 34px;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+  transform: perspective(1px) translateZ(0);
+}
+
+button::before {
+  border: 2px solid $blue-alt-light;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+  border-radius: 70px;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
```diff
@@ -22,3 +24,45 @@ h3 {
...
+}
+
+button:hover::before,
+button:focus::before,
+button:active::before {
+  bottom: -7px;
+  left: 0;
+  opacity: 1;
+  right: -6px;
+  top: 5px;
```
greatislander: Use ems instead of pixels.

assets/css/base/_base.scss
TeddTech: Okay, I will change I chose `#6895c6` it was the closest one in the primary colours. I must not understand the semantics behind the labels Primary vs Secondary. I would love to go over this come Monday.

assets/css/base/_base.scss
TeddTech: Yeah, it does. I think I'll go ahead with that method

assets/css/base/_base.scss
TeddTech: got it :)

assets/css/base/_base.scss
TeddTech: got it

greatislander:
@TeddTech I'm going to finish this. The issue with ems for the button is that ems are relative to the element font size, not the base (`<body>`) font size. So we're trying to have a 2px border (for example) but since the ems are relative to 24px (button font size) instead of 16px (body font size), the math works out to:

0.125 &times; 24px = 3px

The solution is to use rems: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units#Relative_length_units

I'll do this.

PR #96, opened 2020-03-11
feat: SVG logo

greatislander:

## Description

This PR adds an SVG logo which can be styled using the UIO/UIO+ contrast modes feature.

## Steps to test

1. Open the deploy preview in Chrome.
2. Test all contrast modes using UIO+ extension.

**Expected behavior:** Logo changes as expected to match colour schemes.

## Additional information

Not applicable.

## Related issues

Not applicable.


greatislander:
Note — I forked from typography styles branch, so these are the code changes to be reviewed in this PR: https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/96/commits/e86b3ad11c93d233aaa84fce73389676a2e978a3

assets/css/components/_brand.scss
```diff
@@ -0,0 +1,39 @@
+.logo {
+  width: calc(200 / 16 * 1rem);
+}
+
+[class^="fl-theme"] .logo path,
+[class^="fl-theme"] .logo circle,
+[class^="gpii-ext-theme"] .logo path,
+[class^="gpii-ext-theme"] .logo circle {
+  fill: none;
+  stroke: var(--logo-stroke);
```
jobara: @greatislander you can simplify this by just setting stroke to [`currentColor`](https://css-tricks.com/currentcolor/). This will just pick up the current color value which is set through the contrast change. This will eleminate the need to set the color variables, have theme specific styling declarations, and should be more flexible for handling future changes to the contrast themes.

greatislander: Because the SVG image is within a link, using `currentColor` adopts the link colour for the dark grey, light grey, and black on brown colour schemes. Since it works properly for the others, I'm wondering if this is an issue with UIO+?

jobara: Ah I see, well that's not an issue with UIO+ that's because those contrast themes aren't binary and set different colours for links. So, yes in those cases then the colour would be different. 

It's an interesting problem because on one hand the contrast theme is saying that links should be a different colour. On the other hand, I'm not sure you want to treat the site logo the same as the text links on the page.

Curious about your thoughts. Below are a few options I can think of:

- modify the themes so that they don't style SVGs in the same way as other links, at least in terms of colour. Although this would affect other links.
- provide a class or some other mechanism to target certain elements to not style like the rest of the links.
- leave this to the integrator to handle ( i.e. you would have to override them in your styles in this case)

greatislander: My inclination would be to specify the SVG colour (grey for light/dark grey theme and black for black on brown theme) within the stylesheet here. Thoughts? I agree, using the link colour doesn't feel quite right.

jobara: That would be the least prescriptive approach in terms of modifying UIO themes. I guess we can revisit the issue in UIO if we run into this a lot and it because annoying or error prone to handle.

greatislander: Sounds good. One clarification: UIO+ light grey/dark grey body classes are `gpii-ext-theme-gw` and `gpii-ext-theme-gd` respectively. In the UIO [integration guide](https://docs.fluidproject.org/infusion/development/tutorial-userInterfaceOptions/UsingImagesInContrastModes.html), it looks like the equivalent to the dark grey class is `fl-theme-lgdg`. Has anything changed here that I should be aware of on the UIO side?

jobara: This is the [styling](https://github.com/GPII/gpii-chrome-extension/blob/master/extension/stylus/enactors.styl#L13-L60) for the contrast themes used by UIO+. 

This is the [styling](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/css/stylus/utils/Themes.styl#L4-L55) for the contrast themes used by UIO.

The prefs framework/UIO supports an additional theme light-grey-dark-grey (lgdg) which we don't use in UIO+, but you'll probably want to support with the integration of UIO on the site. It won't be affected by the link colour issue we are talking about here though, because it is also a binary theme.

jobara: Seems like the integration guide is a bit out-of-date :(

assets/css/components/_brand.scss
```diff
@@ -7,33 +7,18 @@
...
   --logo-stroke: #{$black};
 }
 
-.fl-theme-gw .logo-element,
-.gpii-ext-theme-gw .logo-element,
-.fl-theme-gd .logo-element,
-.gpii-ext-theme-gd .logo-element {
+.fl-theme-lgdg,
+.gpii-ext-theme-gw,
+.gpii-ext-theme-gd {
```
jobara: I mentioned this in a comment, but it was after you made this commit. UIO also supports `fl-theme-gw` and `fl-theme-gd`. The `fl-theme-lgdg` is a different theme that isn't used in UIO+ but is available in UIO. It is a binary theme though so you won't need to address it separately here, but you will need to address the other two classes I mentioned.

assets/css/abstracts/_variables.scss
```diff
@@ -23,3 +23,12 @@ $blue-alt-light:#8caabb;
...
 $green-pale: #f7fff4;
+
+// UIO
+$yellow: #ff0;
+$white: #fff;
+$grey: #bdbdbb;
+
+// title
+$title-size: 4em;
+$title-weight: semi-bold;
```
TeddTech: I have removed the title variable in another PR

greatislander: Once that PR is merged I will update this.

components/Brand.vue
```diff
@@ -1,8 +1,17 @@
...
 	<div class="brand">
 		<nuxt-link to="/">
-			<img src="~/static/img/We Count logos_colour and bw-01.png" alt="Logo for Project We Count">
+			<WeCountLogo />
 			We Count
 		</nuxt-link>
 	</div>
 </template>
+<script>
+import WeCountLogo from "~/assets/images/logo.svg?inline"
```
TeddTech: Was wondering why you implemented the image this way? as opposed to an img tag or another method

greatislander: If you embed and `<svg>` element, you can style parts of it with CSS. That's what enables the colour changes when the UIO+ theme is changed. SVGs in `` tags can't be styled.

greatislander: Ref:

- https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_and_CSS
- https://css-tricks.com/using-svg/

TeddTech: Interesting! thanks

jobara: If there is an aversion to inlining the SVG, you could look at using `<use>`. I wrote up some research on SVGs for icons and styling them, which you can read on the wiki ( https://wiki.fluidproject.org/display/fluid/Research+SVG+Icons ). That was a while ago, so things may have improved since then. And this will also depend on the browser support we want for the site. Although some of our styling choices for the layout may already be dictating that.

TeddTech: Okay, will check it out

PR #119, opened 2020-03-21
Issue-112: Implement the header layout

cindyli:

## Description
1. Implement the layout for the site header. 
2. Take the chance to adjust the style linting config to set the indentation rule to 4 spaces.

## Steps to test
1. Run `npm run dev`
2. Test the header layout in desktop and mobile browsers.

**Expected behavior:** 
The header style should match [the visual design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A0).


assets/css/layout/_header.scss
```diff
@@ -0,0 +1,149 @@
+// Header styles
+
+header {
+    .brand {
+        a {
+            color: $black;
+            font-size: rem(48);
+            font-weight: normal;
```
TeddTech: @greatislander suggested we use the actual numeric font-weight instead of keywords. For example using `400` instead of `normal`. sass variables for these values are also available in _variable.scss. In this case, you would use `$font-weight-normal` instead of `normal`

stylelint.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
 		"stylelint-order"
 	],
 	"rules": {
+		"indentation": 4,
```
TeddTech: Was wondering why this was changed to 4?

cindyli: The rule of using 4 spaces for indentations is applied in other IDRC projects. I wonder it would be useful to be in sync so developers don't have to change their IDE settings when moving across projects.

Other rules for js files such as indentation using 4 spaces for indentations and having semicolon at the end of lines are also applied in other projects. I lean towards applying the same to this project but let's discuss at today's check-in. Thanks.

TeddTech: Okay sounds good. The rest of this PR lgtm

greatislander: @cindyli Do you use [editorconfig](https://editorconfig.org) in your IDE? This might be a good solution to allow the editor to adopt whatever the code styling conventions are for a given project. I don't have any issue with this change, though.

assets/css/layout/_header.scss
```diff
@@ -0,0 +1,149 @@
...
+            padding: rem(8);
+        }
+
+        button::before {
+            content: none;
+        }
+        // End of overriding the default button styles from _base.scss
+
+        button:hover {
+            background-color: $orange-focus;
```
greatislander: I think the button needs a focus style as well — and the text colour should change on hover/focus (I'd use the same colour as the hover/focus states for the deskop nav links).

assets/css/layout/_header.scss
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+        }
+    }
+}
+
+@media screen and (min-width: 600px) {
+    header {
+        .site-nav {
+            justify-content: space-around;
```
greatislander: I'd suggest setting a `max-width` for this element with `margin-left` and `margin-right` set to auto. On my 27" monitor it looks very strange to have the items spread out this far:

Screen shot of desktop navigation on large screen.

assets/css/layout/_header.scss
```diff
@@ -0,0 +1,149 @@
...
+
+                a {
+                    margin: 0;
+                    padding: rem(32) rem(10);
+                }
+
+                // When the user is in a section (other than the Home page), underline this particular item
+                // in the top level navigation.
+                a.nuxt-link-active {
+                    border-bottom: rem(16) solid $orange-focus;
```
greatislander: The border feels a little heavy to me. Not sure if @lliskovoi has any design input.

Screen shot of menu showing current item indicator.

components/Header.vue
```diff
@@ -1,8 +1,13 @@
 <template>
 	<header>
 		<Brand />
-		<NavBar />
-		<SearchForm />
+		<div class="site-nav">
+			<button id="menuToggleButton" @click="toggleNavMenu()" href="javascript:void(0);" aria-expanded="false">
+				&#9776;&nbsp;Menu
```
greatislander: VoiceOver reads this as 'Button, trigram for heaven menu'! I think we should create an SVG for the menu icon instead of using this character.

components/SearchForm.vue
```diff
@@ -1,11 +1,13 @@
...
-			<label for="search">Search for:</label>
-			<input id="search" type="search" placeholder="Search..." name="s">
-			<button type="submit">
-				Search
-			</button>
+			<input
+				name="s"
+				type="search"
+				placeholder="Search..."
+				title="Enter keywords for a site-wide search"
```
greatislander: I think having both a `title` and an `aria-label` might not be necessary but @lliskovoi might have insight.

cindyli:
All review comments have been addressed. This pull request is ready for another review. Thanks.

assets/css/layout/_header.scss
greatislander: @cindyli The text colour for the focus state is still white, it should be darker.

Screen Shot 2020-03-25 at 11 12 06 AM

assets/css/layout/_header.scss
greatislander: Sorry @cindyli, I think I explained this badly. The header should still be full width but the nav menu/search contaienr should be set to a max width. Right now there is a weird gap on either side, and I think the blue background should extend to full width.

Screen Shot 2020-03-25 at 11 13 26 AM

cindyli:
> I think the breakpoint for switching from mobile nav to desktop nav needs to be increased. Maybe 1024px? Right now we get a bunch of overflow between 600px-100px. Thoughts @cindyli?

Good point. Will make the adjustment.

cindyli:
@TeddTech as now the netlify site is fetching data from a non-production backend. I added a few more views-typed posts into it. The "new and views" page now has more than 1 page and pagination is showing. Could you style the pagination too. Thanks.

cindyli:
All comments have been addressed. This pull request is ready for another look.

Issue #89, opened 2020-03-06
API: split routes for News and Views


greatislander:
## Is your feature request related to a problem? Please describe.

The News page will contain short excerpts linking to external articles/videos/etc. The Views page will contain full length articles written by members of the We Count project. This will necessitate creating separate pages at `/news` and `/views` and populating the content from category feeds from the WordPress API.

## Describe the solution you'd like

Instead of the existing `/news-and-views` page, we need to create separate pages at `/news` and `/views`, populating the content from category feeds from the WordPress API:

- News: https://wecount.inclusivedesign.ca/wp-json/wp/v2/posts?categories=1
- Views: https://wecount.inclusivedesign.ca/wp-json/wp/v2/posts?categories=8

## Describe alternatives you've considered

Not applicable.

## Additional context

Not applicable.

PR #124, opened 2020-03-26
Split news and views

cindyli:

## Description
To address https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/89. It includes:

1. Split the "News and Views" page into two separate pages: "News", "Views";
2. Clicking on a news item redirects the user to the external news;
3. Clicking on a views item brings up an internal page showing its content;
4. Refactored the store to fetch data only once when the website starts;
5. Add "numOfRecPerPage" in config.js to control the number of records shown for one page. This value applies to news, items and search pages.

## Steps to test

1. Run `npm run dev` to start the website;
2. Open the site at http://localhost:3000;
3. The feature described above should match


cindyli:
This pull request is now up to date and ready for a review.

shared/DataFetcher.js
```diff
@@ -0,0 +1,83 @@
...
+		}
+		return results
+	},
+
+	async sitePages () {
+		// According to the Wordpress API for pagination and embedding: https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/,
+		// fetch 100 records per page (the maxium number per page supported by Wordpress) to fasten the query
+		const pageAPI = Config.wpDomain + Config.apiBase + "pages?per_page=100"
+
+		// Fetch records for the first page as well as the number of total pages
```
TeddTech: I believe this comment only applies to posts

shared/DataFetcher.js
```diff
@@ -0,0 +1,83 @@
...
+		const response = await axios.get(`${pageAPI}`)
+		const results = []
+
+		response.data.map(function (onePage) {
+			// The counter is used to navigate from post to post when viewing individual news or views
```
TeddTech: Same as comment above

components/NavBar.vue
```diff
@@ -12,8 +12,11 @@
 		<nuxt-link to="/our-data/">
 			Our Data
 		</nuxt-link>
-		<nuxt-link to="/news-and-views/">
-			News and Views
+		<nuxt-link to="/news/page/1">
```
TeddTech: I believe you do not need the `/page/1`. Just `to="/news/"` should be fine.

components/NavBar.vue
```diff
@@ -12,8 +12,11 @@
...
+			News
+		</nuxt-link>
+		<nuxt-link to="/views/page/1">
```
TeddTech: I believe you do not need the `/page/1`. Just `to="/views/"` should be fine.

components/NavBar.vue
cindyli: Redirecting to `/news/` no longer works because I've removed `/pages/news/index.vue`, which is pretty much a duplicate of `/pages/news/_num/index.vue` with the page number set to 1. Redirecting to `/news/page/1` has the same effect and saves us from maintaining an extra script. 

It's the same reason that "views" link is now pointing to `/views/page/1`.

cindyli:
Thanks for the review, @TeddTech. All comments have been addressed and ready for another look.

components/NewsItem.vue
```diff
@@ -1,18 +1,19 @@
...
-		<figure class="preview-media">
+		<figure v-if="showPreviewImage" class="preview-media">
 			<img v-if="picture" :src="picture" :alt="altTag" class="news-item-img">
 			<WeCountLogo v-else class="news-item-img" />
 		</figure>
 		<h2 class="h3">
-			<nuxt-link :to="'/news-and-views/' + slug">
-				{{ title }}
-			</nuxt-link>
+			<a v-if="isExternalHref" :href="href" v-html="title" target="_blank" />
```
greatislander: I don't think we should use `target="_blank"`. See:

- https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/links-and-new-windows/
- https://css-tricks.com/use-target_blank/

components/NewsItem.vue
```diff
@@ -1,18 +1,19 @@
...
+			<nuxt-link v-else :to="href" v-html="title" />
 		</h2>
+		<div v-html="author" class="author" />
```
greatislander: If the link is to an external resource, this is misleading because it shows who posted the link to our site, not the actual author. I'd hide it if the link is to an external news item.

components/NavBar.vue
greatislander: I'm mildly concerned about this because we end up with 404 (actually, a different error) for `/news/` and `/views/`. I think the convention with paginated content is that visiting the base path shows you the first page. @cindyli is there any way we can display the first page of content at the base URL and all subsequent paginated content at the `/page/<pagenumber>/` route? I understand this might require a duplicate layout but I think it's worth doing.

cindyli:
@greatislander, thanks for the suggestions. The base path access for "/views" and "/news" are added back. I refactored the code a bit so `.vue` files at "/views", "/views/page/$pageNum", "/news" and "/views/page/$pageNum" can share the same template.

All comments have been addressed. Ready for another review. Thanks.

greatislander:
Thanks @cindyli!

Issue #113, opened 2020-03-18
Layout: footer


greatislander:
## Is your feature request related to a problem? Please describe.

Build mobile and desktop styles for three-column footer, including supporting organizations, social media, etc.

## Describe the solution you'd like

See [Figma](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A1).

## Describe alternatives you've considered

Not applicable.

## Additional context

Not applicable.

PR #133, opened 2020-03-30
Issue-113: Layout footer

cindyli:

## Description

To resolve https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/113.

Implement [the footer design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A116)

## Steps to test

1. Run "npm run dev" to start the website and access it at "http://localhost:3000";
2. The footer style should match what's on the design.

## Additional information

As social media accounts for WeCount project haven't been setup, hrefs for social media icons are left empty in this pull request. According to [the IRC discussion with @lliskovoi](http://irc-logs.fluidproject.org/%23fluid-design/%23fluid-design.2020-03-31.log), starting at 2020-03-31T16:36:17, the person who merges this pull request needs to adjust the css to hide the "follow us on social media" section.


assets/css/layout/_footer.scss
```diff
@@ -0,0 +1,158 @@
...
+
+    h2 {
+      font-size: rem(16);
+      font-weight: bold;
+    }
+
+    .contact-info {
+      width: 100%;
+
+      a {
```
greatislander: Insufficient contrast for these links:

Screen Shot 2020-03-30 at 10 12 59 AM

components/ContactInfo.vue
```diff
@@ -3,18 +3,30 @@
...
+		<div>
 			<a :href="'mailto:' + email">{{ email }}</a><br>
 			IDRC<br>
 			205 Richmond Street West<br>
-			<img src="~/static/img/idrc.png">
-		</p>
+			<div class="logo-ocadu-wrapper">
+				<IdrcLogo />
+				<div class="logo-ocadu">
+					<OcaduLogo />
```
greatislander: Should these be linked to the corresponding websites? Whether or not we go that route, they need accessible labels.

components/Funders.vue
```diff
@@ -1,20 +1,20 @@
...
-			<li v-for="f in funders" :key="f.id">
-				<nuxt-link :to="f.href">
-					<img :src="f.logo" :alt="f.logoAlt">
-					{{ f.title }}
-				</nuxt-link>
-			</li>
-		</ul>
+		<p>Supported by Innovation, Science and Economic Development Canada’s Accessible Technology Program.</p>
+		<div>
+			<CanadaLogo />
```
greatislander: Should this be linked to ISED (Accessible Technology Program)? Regardless, it needs an accessible label.

components/SocialMedia.vue
```diff
@@ -1,13 +1,11 @@
...
-		<ul>
-			<li v-for="s in socialMedia" :key="s.id">
-				<nuxt-link :to="s.href">
-					{{ s.title }}
-				</nuxt-link>
-			</li>
-		</ul>
+		<div class="social-media-icons">
+			<div v-for="s in socialMedias" :key="s.id">
+				<nuxt-link :to="s.href" v-html="s.svg" rel="external" />
```
greatislander: These need accessible labels. It might be good to create a `.sr-only` class like this: https://tailwindcss.com/docs/screen-readers/#app

assets/css/layout/_footer.scss
```diff
@@ -0,0 +1,158 @@
+// Footer styles.
+
+footer {
+  background: $navy-light;
+  color: $white;
+  margin-top: rem(30);
+  padding: rem(16) rem(32);
+
+  .footer-content {
```
greatislander: This needs `margin-left: auto` and `margin-right: auto` for large screens:

Screen Shot 2020-03-30 at 10 18 28 AM

cindyli:
@jobara @TeddTech, all review comments have been addressed. This pull request is ready for another review. Thanks.

assets/css/layout/_footer.scss
```diff
@@ -6,90 +6,131 @@ footer {
...
+    .nav-our-data,
+    .nav-inclusion-challenges,
+    .contact-info,
+    .funders {
       a:hover,
       a:focus {
         color: $navy-light;
+
+        svg: {
+          color: $navy-light;
```
greatislander, line 90: Consider using `fill: currentColor` on the SVG: https://css-tricks.com/cascading-svg-fill-color/

cindyli: > Consider using `fill: currentColor` on the SVG: https://css-tricks.com/cascading-svg-fill-color/

Very good point. Following your suggestion of using `currentColor`, I found there's already a similar css set for svg elements: `svg:not([fill]) {fill: currentColor}`. This line of explicitly setting SVG color at hover/focus is unnecessary. I've removed it. Thanks.

assets/config.js
```diff
@@ -13,8 +13,28 @@ export default {
 	loadDbName: "starter_wp",
 	apiBase: "/wp-json/wp/v2/",
 	contactEmail: "wecount@inclusivedesign.ca",
-	socialMedia: [],
-	funders: [],
+	// Using raw SVG content is to work around the issue with dynamically loading and injecting inline SVGs into the template.
```
jobara, line 16: Another approach we can look into for the future is to `<use>`. This will allow you to externally reference an SVG and still be able to style it. However, it does not work in IE 11 and would need a fall back for that. The benefit, other than cleaner looking HTML files, is that you benefit from browser caching.

This [article](https://tympanus.net/codrops/2015/07/16/styling-svg-use-content-css/) has a very detailed explanation of using `<use>` and styling SVGs. Perhaps we should file an issue for this.

components/ContactInfo.vue
```diff
@@ -3,18 +3,36 @@
 		<h2>
 			Contact Us
 		</h2>
-		<p>
+		<div>
```
jobara, line 6: You might want to use the `[<address>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address)` element here instead.

components/ContactInfo.vue
```diff
@@ -3,18 +3,36 @@
...
 			<a :href="'mailto:' + email">{{ email }}</a><br>
 			IDRC<br>
 			205 Richmond Street West<br>
```
jobara, line 9: The address is missing the city and postal code. This was updated in the designs.

components/Footer.vue
```diff
@@ -1,9 +1,56 @@
 <template>
 	<footer>
-		<ContactInfo />
-		<SocialMedia />
-		<Funders />
-		<p>
+		<div class="footer-content">
+			<div class="nav-home-about">
+				<div class="nav-link-wrapper">
```
jobara, line 5: What is the purpose of `nav-link-wrapper`? Can we get the same styling without the wrapper? Is this needed because of the sub-nav that isn't implemented yet? If we do need it, is there a more descriptive name that could be used?

components/Footer.vue
```diff
@@ -1,9 +1,56 @@
 <template>
 	<footer>
-		<ContactInfo />
-		<SocialMedia />
-		<Funders />
-		<p>
+		<div class="footer-content">
+			<div class="nav-home-about">
```
jobara, line 4: We should consider making these or perhaps all of the navigation here into a list, that way a screen reader user would know how many items are there. We should also consider wrapping them all in a `<nav>` element so that the user knows they are for navigation. Maybe check with @lliskovoi about her thoughts.

assets/config.js
cindyli: Umm... By reading the article, I don't think it will help with solving the issue with dynamically loading SVGs. The syntax to use `<use>` is:

```
<symbol id="my-icon" viewBox="0 0 30 30">
	<!-- icon content / shapes here -->
</symbol>

<use xlink:href="#my-icon" x="100" y="300" />
```
It requires the icon content to be preloaded into `<symbol>`. We'll hit the same loading issue here.

Please correct me if my understanding is wrong. Thanks.

components/Footer.vue
```diff
@@ -1,9 +1,56 @@
...
+				<div class="nav-link-wrapper">
+					<nuxt-link to="/" aria-label="Goto Home">
```
jobara, line 6: I think "Goto" should be "Go to"

components/Footer.vue
cindyli: `nav-link-wrapper` is to add space in between the "home" link and "about" link. This space is also needed when the secondary nav items are in place. My another choice is to remove `nav-link-wrapper` and add padding between link directly. But it will enlarge the link itself and focus/hover styling for links will highlight the whole space, which I feel is less ideal.

assets/config.js
jobara: `<symbol>` is defined in the svg sprite. It is used to group the svg shapes so they can be referenced externally by `<use>`. The href in `<use>` would specify the svg sprit file and the # would reference the particular symbol. I've written this up on the [Research SVG Icons](https://wiki.fluidproject.org/display/fluid/Research+SVG+Icons) wiki page when I was looking into this stuff several years ago. It seems like we're getting to a point where support in most browsers could actually make this a viable solution. At the time I was researching this there were issues with styling elements.

components/Footer.vue
cindyli: @jobara, what do you think about renaming "nav-link-wrapper" to "nav-link-block"?

components/Footer.vue
lliskovoi: I agree. It should be a list with secondary items as nested lists, and should be wrapped in a <nav> with aria-label="footer" or something along those lines.

components/Footer.vue
jobara: See @lliskovoi's comments from https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/133#discussion_r401686538. So there will be block level element containers with the `<li>` that can be used instead. Also we can either use a class like you suggested, or modify the styling to just target the second element. for example you could do something like `.nav-home-about > li  ~ li { margin-top: 2rem }`; which I believe will target all the `<li>` elements that are a direct child of `.nav-home-about` and siblings to the first `<li>`.

components/Funders.vue
```diff
@@ -1,20 +1,23 @@
...
-				</nuxt-link>
-			</li>
-		</ul>
+		<p>Supported by Innovation, Science and Economic Development Canada’s Accessible Technology Program.</p>
+		<div>
+			<a href="https://www.ic.gc.ca/eic/site/118.nsf/eng/home">
+				<CanadaLogo />
+				<span class="screen-reader-only">Accessible Technology Program</span>
+			</a>
+		</div>
```
jobara, line 9: This may be a question for @lliskovoi. Should the Paragraph be included in the link, or maybe the Program's name. The link is currently from the Canada logo, and also includes some AT only text. 

Screen Shot 2020-04-01 at 11 17 34 AM

jobara: @cindyli the space between the logo and the text seems wider in the implementation than in the mockups. In the mockups they are 13px apart. Which would be about 0.8125 rems (if 16px base font size).

Implementation:
Screen Shot 2020-04-01 at 11 19 40 AM

Mockups:
Screen Shot 2020-04-01 at 11 20 58 AM

lliskovoi: @jobara Do you mean as a single focusable element? Would it look weird if only "Innovation, Science and Economic Development Canada's Accessible Technology Program" was included? Don't think we should include "Supported by" in the link text if we combine with the image. I would suggest either the logo only be a link, or just the "Innovation..." text.

components/Footer.vue
cindyli: Css selectors work great. Thanks, @jobara.

cindyli:
I believe all review comments have been addressed. This pull request is ready for another review.

jobara:
@cindyli the SVGs look too small in the mobile view, with the exception of the Canada logo.
For example the Social media icons are 2.5rem in the implementation but 4.875rem in the mockup.

It may also be the reason, or part of the reason, why the IDRC logo doesn't look centred.

Implementation:
Screen Shot 2020-04-01 at 1 06 45 PM

Mockups:
Screen Shot 2020-04-01 at 1 07 19 PM

jobara:
@cindyli and @lliskovoi there are a couple issues with the Social media icons on hover. 

1. The stroke expands beyond the border so it clips on edges. 
2. The stroke is thinner than in the mockups.

Screen Shot 2020-04-01 at 1 12 46 PM

cindyli:
@jobara thanks for the effective help on improving SVG focus/hover styles. All comments are addressed and ready for another review.

assets/config.js
cindyli: I will file an issue for this when this pull request is merged so the issue can use links to the code in the master branch.

jobara:
@cindyli this looks good to me. However, I didn't get a chance to really look over the grid layout for the footer and nav. Maybe @greatislander could review that part.

components/Footer.vue
```diff
@@ -1,9 +1,59 @@
...
-		<SocialMedia />
-		<Funders />
-		<p>
+		<div class="footer-content">
+			<nav aria-label="Footer" class="footer-nav">
+				<ul>
+					<li>
+						<nuxt-link to="/" aria-label="Go to Home">
+							Home
+						</nuxt-link>
```
greatislander, line 9: I don't think the `aria-label` is necessary. A link implicitly means "Go to [somewhere]". Also, overriding the link text with an `aria-label` means that a non-English user who is using Google Translate to translate the page will not get a translated version of the link text (`aria-label` doesn't get translated by translation tools: https://adrianroselli.com/2019/11/aria-label-does-not-translate.html).

components/Pagination.vue
```diff
@@ -7,33 +7,33 @@
 				</nuxt-link>
 			</li>
 			<li v-if="currentPageNum > 2">
-				<nuxt-link :to="firstLink" class="pagination-link" aria-label="Goto page 1">
+				<nuxt-link :to="firstLink" class="pagination-link" aria-label="Go to page 1">
```
greatislander, line 10: I also wonder about this `aria-label` use. I can see the the benefit of providing "Page" context for the numeric link text but I feel that this may be better-served by including visually-hidden text within the link itself (using the `screen-reader-only` class).

assets/css/layout/_footer.scss
```diff
@@ -0,0 +1,242 @@
...
+    .footer-nav {
+      display: none;
+    }
+
+    .footer-nav,
+    .contact-info,
+    .funders {
+      a:hover,
+      a:focus {
+        color: $navy-light;
```
greatislander, line 92: Comparing the hover/focus states of the footer navigation to the header navigation, I wonder if using a larger box shadow might be good. I find the links look tiny surrounded by a lot of blue space.

### Header

Screen Shot 2020-04-03 at 9 25 33 AM

### Footer

Screen Shot 2020-04-03 at 9 25 15 AM

assets/css/layout/_footer.scss
```diff
@@ -0,0 +1,242 @@
...
+      grid-template-columns: 3.8fr 1.2fr;
+      padding: rem(32) rem(48);
+
+      .footer-nav {
+        display: block;
+
+        a {
+          color: $green-light;
+          font-weight: $font-weight-semibold;
+        }
```
greatislander, line 154: There is no current page indicator on the footer link styles. The header navigation adds a `border-bottom` to links with the `.nuxt-link-active` class.

I also noticed that the Home link in the footer has a `.nuxt-link-active` class on all pages. This seems like a bug.

Lastly, the `.nuxt-link-active` class doesn't provide any info to assistive technology users. The correct way of indicating that a link represents the current page is using `aria-current="page"`. I'm wondering if the `<NuxtLink>` component supports this?

components/Footer.vue
lliskovoi: I agree, especially if it's clear that this is a pagination navigation section, the function of the number links is clear on its own.

assets/css/layout/_footer.scss
cindyli: > There is no current page indicator on the footer link styles. The header navigation adds a `border-bottom` to links with the `.nuxt-link-active` class.

It's decided at yesterday's meeting that using `aria-current="page"` for the active link on the footer is enough and doesn't need to add a visual indicator.

cindyli:
All review comments have been addressed and ready for another look.

cindyli:
Good catch, @greatislander, and the solution article works great. Thanks.

Ready for another review.

PR #122, opened 2020-03-26
Style homepage content

TeddTech:
* [x] This pull request has been linted by running `npm run lint` without errors
* [ ] This pull request has been tested by running `npm run dev` and reviewing affected routes
* [ ] This pull request has been built by running `npm run generate` without errors
* [x] This isn't a duplicate of an existing pull request

## Description
**PR #120 NEEDS TO BE MERGED FIRST**
Build home page content area.

## Steps to test

1. Run npm run dev
2. Test the header layout in desktop and mobile browsers.

**Expected behavior:** 

This should match the [visual design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A0).

## Related issues

resolves #114

PR #134, opened 2020-04-01
Style page content

### LINKED ISSUE #115: Layout: page content

Build mobile and desktop styles for page content.

See [Figma](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A1).

Not applicable.

## Additional context

We will need to build the in-page table of contents which appears in the sidebar, dynamically adding ids and links to level-two headings.

TeddTech:

## Description

Built mobile and desktop styles for page content.

## Steps to test

1. Run "npm run dev" to start the website and access it at "http://localhost:3000"
2. Page content style should match what's on the design.

**Expected behavior:** 

This should match the [visual design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A0).

## Additional information

On line `9-23` of `components/Aside.vue`, `<li> `are hard coded to show an example of how styles will look. This will be replaced by the commented out html code above which generates table of contents dynamically.

## Related issues

resolves #115


cindyli:
When the nav title is long, the text should wrap. Right now, the long title pushes the main content away. 

Screen Shot 2020-04-01 at 11 17 41 AM

cindyli:
The gap between the side section and the main content is too narrow.

The background image is not in place.

TeddTech:
> The gap between the side section and the main content is too narrow.
> 
> The background image is not in place.

The background image is taken care of in PR #122

assets/css/abstracts/_variables.scss
```diff
@@ -39,3 +39,6 @@ $orange-focus: #fadc90;
 $yellow: #ff0;
 $white: #fff;
 $grey: #bdbdbb;
+
+// Images
+$img-size-medium: rem(262); // need to decide and set other size variables
```
jobara, line 44: I'm not sure what this comment means. Is it still needed?

components/Aside.vue
```diff
@@ -1,16 +1,32 @@
...
+				<nuxt-link to="#title3">
+					title3
+				</nuxt-link>
 			</li>
 		</ul>
 	</aside>
 </template>
 
 <script>
+// The above <li> are coded to show an example of how styles will look. This will be replaced by the commented out html code above.
```
jobara, line 29: I think you should put the commented out code with the comment. Also I'm not sure why the commented out code isn't just being used. Please explain that in the comment.

assets/css/abstracts/_variables.scss
TeddTech: Yes, we need to decide on standard sizes for all types of images.

``` css
// Images
$img-size-small: ?;
$img-size-medium: rem(262);
$img-size-large: ?;
```

I will bring this up in the check-in today.

components/Aside.vue
TeddTech: This was a personal preference. I don't really like how commented out HTML looks. Changing now

TeddTech:
Ready for review

jobara:
@TeddTech I'm getting the following error when running `npm run dev`

```bash
 ERROR  [Vue warn]: Property or method "headers" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.
```

TeddTech:
Ready for review.

components/Aside.vue
```diff
@@ -1,10 +1,26 @@
 <template>
 	<aside>
 		<ul>
-			<li v-for="h in headers" :key="h.id">
+			<!-- <li v-for="h in headers" :key="h.id">
 				<nuxt-link :to="h.href">
 					{{ h.title }}
 				</nuxt-link>
+			</li> -->
+			<!-- The above <li> below are hard coded to show an example of how styles will look. This will be replaced by the commented out html code above. -->
```
jobara, line 9: I think there is a mistake in the comment. I think "above" shouldn't be at the beginning of the sentence.

jobara: Maybe make this a TODO as well.

components/Aside.vue
```diff
@@ -1,10 +1,26 @@
...
+				<nuxt-link to="#title2">
+					title2
+				</nuxt-link>
+			</li>
+			<li>
+				<nuxt-link to="#title3">
+					title3
+				</nuxt-link>
 			</li>
 		</ul>
```
jobara, line 25: I think this should all be placed inside a `<nav>` element. @lliskovoi shared a [link](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role) in the fluid-design channel that discusses using the `<nav>` element and in particular when there are multiple `<nav>` elements on the page.

jobara:
@TeddTech It looks like the in page navigation sticks to the left edge. I think we'll probably want to cap that on really wide screens to line up with the left edge of where the we count logo stops. @lliskovoi could you confirm how you'd like this implemented?

Screen Shot 2020-04-01 at 2 18 25 PM

jobara:
@TeddTech at 1024px width screens, the in page nav starts pushing into the text, causing it to wrap around the navigation.

Screen Shot 2020-04-01 at 2 21 22 PM

jobara:
The wireframes for the Mandate section has the image to the right of all the text including the heading. However, in the implementation it is at the bottom right corner or the section.

Screen Shot 2020-04-01 at 2 23 01 PM

TeddTech:
> @TeddTech at 1024px width screens, the in page nav starts pushing into the text, causing it to wrap around the navigation.
> 
> Screen Shot 2020-04-01 at 2 21 22 PM

@cindyli @jobara @greatislander @lliskovoi Yesterday we mentioned changing the semantics of the page so that the main content of the page is wrapped around a `<main>` element and it was decided this issue will be discussed further and decided next check-in. 

To solve this issue of text wrapping around navigation I have to put the `<article>` element around another element (`<main>`) and set `display: flex` to that outer element.

I am writing this to ask to go ahead and change the code so that a `<main>` element is wrapping around the main content of the page. The structure I am talking about will look like this for the home page:

Screen Shot 2020-04-02 at 11 26 50 AM

OR THIS: 

Screen Shot 2020-04-02 at 11 26 17 AM

cindyli:
> I am writing this to ask to go ahead and change the code so that a `<main>` element is wrapping around the main content of the page. 

This structure makes sense to me. Other than <article> elements, I feel the pagination displayed on some pages should be included in the <main> too.

However, before we revisit the semantic structure of the entire page including header/footer, it might be easier to use <div> instead of <main>, and leave a TODO comment for the <div>.

greatislander:
@cindyli It looks like some of your comment is missing because it uses HTML. Can you wrap those in backticks (`) so the tags don't disappear?

cindyli:
> Can you wrap those in backticks (`) so the tags don't disappear?

Thanks for the reminder. Fixed.

greatislander:
I agree with @cindyli's point about using a `<div>` for now and leaving a TODO comment. I think taking a look at the overall page semantics would be good. My starting point is always:

```html
<body>
	<header role="banner">
		<a rel="home" href="/">Site Title</a>
		<nav aria-label="primary navigation">
			<!-- primary navigation -->
		</nav>
	</header>
	<main>
		<-- main content, probably in an <article> -->
	</main>
	<footer role="contentinfo">
		<nav aria-label="secondary navigation">
			<!-- secondary navigation -->
		</nav>
	</footer>
</body>
```

Semantically, I think that you're correct, @TeddTech, in placing the `<aside>` outside of the `<main>` tag as it is a complementary landmark (see: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/complementary.html)

TeddTech:
Ready for review

cindyli:
On "tools" and "our data" pages, the side menu shows up in the middle of the page (see attached image). It should be to the left.

Screen Shot 2020-04-03 at 9 55 15 AM

layouts/default.vue
```diff
@@ -1,17 +1,22 @@
 <template>
 	<div id="defaultContainer">
 		<Header />
-		<nuxt />
+		<main>
```
greatislander, line 4: @TeddTech In my comment I said:

> I agree with @cindyli's point about using a `<div>` for now and leaving a TODO comment.

I also said:

> Semantically, I think that you're correct in placing the `<aside>` outside of the `<main>` tag as it is a complementary landmark…

However, you used `<main>` instead of a `<div>` and you put the `<aside>` inside the `<main>`. This is not semantically correct, and as per @cindyli's comment (https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/134#issuecomment-608451020), this doesn't address some of these layout issues.

greatislander:
@cindyli Ready for review — before we merge we should remove the test content in `<Aside>` component.

TeddTech:
> @cindyli Ready for review — before we merge we should remove the test content in `<Aside>` component.

Also, I believe #122 needs to merged first. Right now we are just waiting for the hovering styling for the cards. After that, it should be good to merge as well.

assets/css/components/_aside.scss
```diff
@@ -0,0 +1,48 @@
+aside {
+  display: none;
+}
+
+@media screen and (min-width: 1024px) {
+  aside {
+    display: inherit;
+    margin-left: auto;
```
cindyli, line 8: `margin-left` is no longer needed as its parent wrapper uses the grid display.

Issue #170, opened 2020-04-08
Auto generate navigation menus based on the API response


cindyli:
## Is your feature request related to a problem? Please describe.

Right now, main navigation menus on the header and the footer are hardcoded.

## Describe the solution you'd like

The main navigation menu should be auto generated based on the response from the WordPress API: https://wecount-dev.inclusivedesign.ca/wp-json/wp/v2/pages

## Describe alternatives you've considered

N/A.

## Additional context

N/A.

greatislander:
Pages can be assigned a `menu_order` attribute in WordPress. It's an integer value. See the "Order" field in the right column (exposed in the REST API here: https://wecount-cms.inclusivedesign.ca/wp-json/wp/v2/pages/118):

<img width="1114" alt="Screen Shot 2020-04-09 at 3 25 27 PM" src="https://user-images.githubusercontent.com/605361/78942173-5cceae80-7a76-11ea-8959-fa32f996e3a1.png">

A strategy for the navigation could be using this field to indicate which pages should be included in the menu, with the following logic:

1. Only include pages in the menu if they have a `menu_order` value > 0.
2. Sort the pages by `menu_order` value in ascending order.

greatislander:
This would mean that News and Views would need to be appended to the end of the menu, assuming that the other pages appeared first (they don't correspond to pages in WordPress). Maybe this idea still needs work.

greatislander:
https://wecount-cms.inclusivedesign.ca/wp-json/wp/v2/pages?order=asc&orderby=menu_order

PR #196, opened 2020-04-14
Issue-170: Auto generate navigation menus based on the API response.

cindyli:

## Description

To resolve https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/170.

Navigation menus on the header and footer are auto generated based on the WordPress API response.

## Steps to test

1. The navigation menu on the header shows 4 items: About, Inclusion Challenges, News, Views;
2. Remove `style="display: none;"` that hides the nav menu on the footer;
3. The nav menu on the footer shows 5 items: Home and 4 items described in 1.

The footer css has been adjusted to work with the current 5 menu items instead of 7 that was expected in the design. The footer css needs to be revisited every time when a menu item is added or removed.

**Expected behavior:** 

The navigation menu should look the same as what's in the current dev branch.

### DIFF
--- a/shared/Utils.js
+++ b/shared/Utils.js
@@ -6,11 +6,11 @@ export default {
 	},
 
 	/**
-	 * The complete Triforce, or one or more components of the Triforce.
-	 * @typedef {Object} menuItem
-	 * @property {String} slug - The slug of the corresponding page that the menu item lands on.
-	 * @property {String} title - The title of the corresponding page that the menu item lands on.
-	 * @property {String} href - The href to direct the menu item to its corresponding page.
+	 * generateNavMenu() returns an array of menu items. The structure of a menu item is:
+	 * @typedef {Object} MenuItem
+	 * @property {string} slug - The slug of the corresponding page that the menu item lands on.
+	 * @property {string} title - The title of the corresponding page that the menu item lands on.
+	 * @property {string} href - The href to direct the menu item to its corresponding page.
 	 */
 
 	/**
@@ -18,14 +18,17 @@ export default {
 	 * The retun menu array will:
 	 * 1. Exclude all pages that have menu_order === 0;
 	 * 2. Append "News" and "Views" menu items to the end.
-	 * @param {Array<Object>} sitePages - All menu pages sorted by menu_order.
-	 * @param {Boolean} includeHomeItem - A flag indicating if the home menu item should be included.
-	 * @return {Array<menuItem>} An array of object including the slug, menu title and href.
+	 * @param {array<Object>} sitePages - All menu pages sorted by menu_order.
+	 * @param {boolean} includeHomeItem - A flag indicating if the home menu item should be included.
+	 * @return {array<MenuItem>} An array of object including the slug, menu title and href.
 	 */
 	generateNavMenu (sitePages, includeHomeItem) {
 		const menuItems = [];
 		const homeMenuItem = [];
-		const itemsToAppend = [{
+		// Since "News" and "Views" don't have landing pages defined at the WP back-end, these menu items need to
+		// be explicitly appended to the nav menu. Using "append" is because they are the last 2 nav elements according
+		// to the design.
+		const extraMenuItemsToAppend = [{
 			slug: "news",
 			title: "News",
 			href: "/news"
@@ -45,7 +48,7 @@ export default {
 			}
 
 			// Exclude pages that have menu_order === 0, which indicates the page should not be displayed on the header nav menu.
-			if (oneSitePage.menu_order) {
+			if (oneSitePage.menu_order !== 0) {
 				menuItems.push({
 					slug: oneSitePage.slug,
 					title: oneSitePage.title,
@@ -54,6 +57,6 @@ export default {
 			}
 		});
 
-		return [...homeMenuItem, ...menuItems, ...itemsToAppend];
+		return [...homeMenuItem, ...menuItems, ...extraMenuItemsToAppend];
 	}
 };
--- a/shared/Utils.js
+++ b/shared/Utils.js
@@ -40,6 +40,9 @@ export default {
 
 		sitePages.forEach((oneSitePage) => {
 			if (includeHomeItem && oneSitePage.slug === "home" && oneSitePage.menu_order === 0) {
+				// The title "Home" is hardcoded based on the slug is because the title received from the API response is
+				// `Creating an inclusive data ecosystem`, which doesn't provide any hint that it's a home page. And it also
+				// doesn't match what's on the design.
 				homeMenuItem.push({
 					slug: oneSitePage.slug,
 					title: "Home",
--- a/shared/Utils.js
+++ b/shared/Utils.js
@@ -6,17 +6,23 @@ export default {
 	},
 
 	/**
+	 * The complete Triforce, or one or more components of the Triforce.
+	 * @typedef {Object} menuItem
+	 * @property {String} slug - The slug of the corresponding page that the menu item lands on.
+	 * @property {String} title - The title of the corresponding page that the menu item lands on.
+	 * @property {String} href - The href to direct the menu item to its corresponding page.
+	 */
+
+	/**
 	 * Generate the navigation menu based on the response from the Wordpress API.
 	 * The retun menu array will:
-	 * 1. Exclude all pages whose menu_order === 0;
+	 * 1. Exclude all pages that have menu_order === 0;
 	 * 2. Append "News" and "Views" menu items to the end.
 	 * @param {Array<Object>} sitePages - All menu pages sorted by menu_order.
 	 * @param {Boolean} includeHomeItem - A flag indicating if the home menu item should be included.
-	 * @return {Array<Object>} An array of object including the slug, menu title and href.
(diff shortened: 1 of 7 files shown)


shared/Utils.js
```diff
@@ -3,5 +3,50 @@
...
+	 * The retun menu array will:
+	 * 1. Exclude all pages whose menu_order === 0;
+	 * 2. Append "News" and "Views" menu items to the end.
+	 * @param {Array<Object>} sitePages - All menu pages sorted by menu_order.
+	 * @param {Boolean} includeHomeItem - A flag indicating if the home menu item should be included.
+	 * @return {Array<Object>} An array of object including the slug, menu title and href.
+	 */
+	generateNavMenu (sitePages, includeHomeItem) {
+		// Don't include the menu item for "home" by default
+		includeHomeItem = includeHomeItem === undefined ? false : includeHomeItem;
```
jobara, line 19: I'm not sure you need this, as `undefined` is already falsey.

shared/Utils.js
```diff
@@ -3,5 +3,50 @@
...
+		sitePages.forEach((oneSitePage) => {
+			if (includeHomeItem && oneSitePage.slug === "home") {
+				homeMenuItem.push({
+					slug: oneSitePage.slug,
+					title: "Home",
+					href: "/"
+				});
+			}
+
+			if (oneSitePage.menu_order > 0) {
```
jobara, line 41: what does `oneSitePage.menu_order` contain? I'm not really sure why it would ever contain a negative value. And it doesn't look like you're using the order to sort the menu items.

shared/Utils.js
```diff
@@ -3,5 +3,50 @@
...
+	 * @param {Array<Object>} sitePages - All menu pages sorted by menu_order.
+	 * @param {Boolean} includeHomeItem - A flag indicating if the home menu item should be included.
+	 * @return {Array<Object>} An array of object including the slug, menu title and href.
+	 */
+	generateNavMenu (sitePages, includeHomeItem) {
+		// Don't include the menu item for "home" by default
+		includeHomeItem = includeHomeItem === undefined ? false : includeHomeItem;
+		const menuItems = [];
+		const homeMenuItem = [];
+		const itemsToAppend = [{
```
jobara, line 22: I think this needs a more descriptive name. I'm not sure why these menu items are hard coded and not included with `sitePages`.

shared/Utils.js
```diff
@@ -3,5 +3,50 @@
...
+		}];
+
+		sitePages.forEach((oneSitePage) => {
+			if (includeHomeItem && oneSitePage.slug === "home") {
+				homeMenuItem.push({
+					slug: oneSitePage.slug,
+					title: "Home",
+					href: "/"
+				});
+			}
```
jobara, line 39: I wonder if you can simplify things here by using [unshift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift).

It seems that `oneSitePage.slug === "home"` will happen at the same time as `oneSitePage.menu_order < 0` otherwise you'd run the risk of adding the home page twice, and I suppose broken in the second case. Instead of pushing to `homeMenuItem` you just unshift to `menuItems` which would put it at the front of the list.

shared/Utils.js
```diff
@@ -3,5 +3,50 @@
...
+	},
+
+	/**
+	 * Generate the navigation menu based on the response from the Wordpress API.
+	 * The retun menu array will:
+	 * 1. Exclude all pages whose menu_order === 0;
+	 * 2. Append "News" and "Views" menu items to the end.
+	 * @param {Array<Object>} sitePages - All menu pages sorted by menu_order.
+	 * @param {Boolean} includeHomeItem - A flag indicating if the home menu item should be included.
+	 * @return {Array<Object>} An array of object including the slug, menu title and href.
```
jobara, line 15: For jsdoc you could setup a [typedef](https://jsdoc.app/tags-typedef.html) which would describe the object being returned. Which you'd then use in place of Object above. That's a way to provide more details on what goes into a specific type of Object.

shared/Utils.js
cindyli: `oneSitePage.menu_order` could be 0 or non-zero. I will change the condition to `oneSitePage.menu_order !== 0` and add a code comment to reduce the confusion.

The sort is performed via [the fetch from WP API](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/auto-generate-nav-menu/shared/DataFetcher.js#L61) with `order=asc&orderby=menu_order`.

shared/Utils.js
cindyli: "News" and "Views" doesn't have a page defined in the WP back-end. Their page contents are populated at the WeCount website based on posts published in these categories.

shared/Utils.js
cindyli: Good catch. To avoid adding the home page twice, I will modify the condition of this block to `includeHomeItem && oneSitePage.slug === "home" && oneSitePage.menu_order === 0` so that the home page will only be added to `homeMenuItem` when it isn't included in the nav menu. Please let me know if you notice a flaw with this solution. Thanks.

shared/Utils.js
jobara: @cindyli, I think you could write `if(oneSitePage.menu_order)` or maybe I'm mistaking what you're saying.

shared/Utils.js
jobara: Do you need to check both `menu_order` and `slug`. Will home always have a `menu_order` that is 0?

cindyli: Based on the current design that excludes the home page from the navigation menu on the header, `menu_order` for the home will always be 0, unless there's a design change.

shared/Utils.js
cindyli: Sure, I will change it to `if (oneSitePage.menu_order)` although my personal preference is to be explicit about what exact condition `if` should handle.

jobara: Ok, that's reasonable for keeping it as you suggested.

shared/Utils.js
```diff
@@ -3,5 +3,57 @@
 export default {
 	stripHtmlTags (inputString) {
 		return inputString.replace(/<\/?[^>]+(>|$)/g, "");
+	},
+
+	/**
+	 * The complete Triforce, or one or more components of the Triforce.
```
jobara, line 9: This should be a specific description of the type you are defining. I think you probably got this from the link I sent, but it was constructed as an example which I believe is actually a reference to a [video game](https://en.wikipedia.org/wiki/Triforce). Although it is somewhat fitting here as well.

shared/Utils.js
```diff
@@ -3,5 +3,57 @@
...
+	 * @typedef {Object} menuItem
```
jobara, line 10: I think these are supposed to be HeadsUp camelcase. Although I'm not 100% sure of that.

shared/Utils.js
jobara: Maybe refer to these as dynamic pages or something? Is there a way to source the reference from outside of this function and pass it in? That way if we add/remove/modify them we won't have to touch this function.

cindyli: This function is called to generate nav menus on the header and the footer. This logic of appending "News" and "Views" is applicable to both places. I think the goal of this function is to be able to deal with this use case at one common place. Passing in these items as an argument would defeat this purpose.

cindyli:
All comments have been addressed except [this one](https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/196#discussion_r408890496) in discussion. @jobara, please let me know your thoughts about my reply. Thanks.

shared/Utils.js
jobara: I guess what I meant is that if we convert Views for example to be a static page, we'll also have to update this function to remove it. Or if we change the Title. This function will have to be updated. I guess that last point goes for Home as well.

If these don't seem like issues that will come up, or are not things we need to deal with now, then we can punt the discussion for later.

cindyli: When we convert News and Views to static pages, these menu items still need to be appended. Creating static page content doesn't change the fact that these pages don't have landing pages defined in WP. 

In terms of the title, yes, if the design changes the title, this function needs to be adjusted. Since there isn't an API fetch telling us what these titles should be, hardcoding them here is what I can come up with. At least only this function needs to be adjusted when there's a title change.

About the title for "Home", what I received from API for this title is `Creating an inclusive data ecosystem`, which doesn't give any hint that it's a home page. And it also doesn't match what's on the design. That's why the title is hardcoded to "Home" based on its slug.

store/actions.js
```diff
@@ -23,5 +23,9 @@ export default {
 			const sitePages = await DataFetcher.sitePages();
 			context.commit("setSitePages", sitePages);
 		}
+	},
+
+	nuxtServerInit ({ dispatch }) {
```
TeddTech, line 28: Thank you for this @cindyli !

shared/Utils.js
jobara: Let's punt this for now. Might be useful to migrate some of this discussion into comments in the code. I'm thinking the issue with `Home` would be a good idea to leave a comment about.

cindyli:
All comments have been addressed. This pull request is ready for another look.

PR #202, opened 2020-04-21
Side menu functionality

### LINKED ISSUE #169: Improve the side menu when both the content and the design are ready

The side menu is initially designed to display secondary navigation items (h2s) to allow the jump to each section in the article. The style of it has been implemented based on the mock data. Improving it to a fully functioning feature is waiting on 2 materials:

1. Contents with h2s. When contents are available, they need to be parsed to extract h2s for display;
2. "News" and "Views" pages will not have h2s but only have the title in h1. @lliskovoi will come up with a design for these pages.

See above.

N/A.

## Additional context

N/A.

TeddTech:

## Description
Display secondary navigation items (h2s) on the left side of the page to allow the jump to each section in the article.

## Steps to test

1. run `npm run generate`
2. run `npx serve dist`
3. Go to the about page and inspect/test aside component
3. Go to the view articles and inspect/test aside component

**Expected behavior:** 

Links in the aside component should jump to section of page with the same title for the h2 header.

## Related issues

resolves #169


shared/SideMenu.js
```diff
@@ -0,0 +1,29 @@
...
+			h2List[i].setAttribute("id", `${h2List[i].textContent.replace(/\s+/g, "-")}`);
+		}
+		return jsdom.serialize();
+	},
+	getHeaderList (content) {
+		const JSDOM = require("jsdom").JSDOM;
+		const jsdom = new JSDOM(content);
+		const { window } = jsdom;
+
+		let headerList = [...window.document.getElementsByTagName("h2")];
```
cindyli, line 20: `window.document.getElementsByTagName("h2")` should already return an array. Is it necessary to reprocess it with `[...h2Array]`?

layouts/default.vue
```diff
@@ -40,6 +45,18 @@ export default {
 	},
 	methods: {
 		setAriaCurrent () {
+			// Logic for sideMenu
```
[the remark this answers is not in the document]
TeddTech: That's what I tried initially but it did not work. May we pair up and try to resolve this.

shared/SideMenu.js
TeddTech: Yes, it is necessary. `window.document.getElementsByTagName("h2")` returns a collection. You cannot do array methods like `.map` on collections only Arrays.

colinbdclark: Couldn't you just use [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)?


```
let h2Collection = window.document.getElementsByTagName("h2");
return Array.from(h2Collection, function (header) {
     return {
          title: header.textContent,
          href: `#${header.textContent.replace(/\s+/g, "-")}`
     };
});
```

TeddTech: Yes, I believe I could. should I change the code to use it?

assets/css/layout/_layout.scss
```diff
@@ -95,7 +95,7 @@ figcaption {
 
 	.content-wrapper {
 		display: grid;
-		grid-template-columns: 1fr minmax(75ch, 130ch) 1fr;
+		grid-template-columns: 1fr minmax(75ch, 130ch) 23ch;
```
greatislander, line 98: What's the rationale for this change? It makes the right column a fixed width, which leads to this kind of weirdness on large screens:

Screen Shot 2020-04-22 at 8 43 50 AM

layouts/default.vue
```diff
@@ -1,26 +1,24 @@
...
 import Header from "~/components/Header";
-import Aside from "~/components/Aside";
 import Footer from "~/components/Footer";
 export default {
 	components: {
 		Header,
-		Aside,
 		Footer
 	},
+	data () {
```
cindyli, line 17: This `data` block can be removed as `<Aside>` has been moved out of this vue file.

nuxt.config.js
```diff
@@ -101,9 +101,15 @@ export default {
...
-		*/
-		extend (config, ctx) {
+				** You can extend webpack config here
+				*/
+		extend (config, { isDev, isClient }) {
+			if (isClient) {
+				config.node = {
+					fs: "empty",
+					child_process: "empty"
+				};
```
greatislander, line 111: Please document this in a comment.

package.json
```diff
@@ -34,8 +34,13 @@
     "@nuxtjs/dotenv": "^1.4.0",
     "@nuxtjs/pwa": "^3.0.0-beta.20",
     "@nuxtjs/svg": "^0.1.7",
+    "child_process": "^1.0.2",
+    "fs": "0.0.1-security",
```
greatislander, line 38: How are these used?

package.json
```diff
@@ -34,8 +34,13 @@
...
+    "jsdom": "^16.2.2",
+    "net": "^1.0.2",
```
greatislander, line 40: How is this used?

package.json
```diff
@@ -34,8 +34,13 @@
...
     "nuxt": "^2.12.2",
-    "sanitize.css": "^11.0.0"
+    "sanitize.css": "^11.0.0",
+    "tls": "0.0.1"
```
greatislander, line 43: How is this used?

nuxt.config.js
```diff
@@ -101,9 +101,15 @@ export default {
 	*/
 	build: {
 		/*
-		** You can extend webpack config here
-		*/
-		extend (config, ctx) {
+				** You can extend webpack config here
```
cindyli, line 104: Please fix the indentations on line 104-105. Interesting that the linting script doesn't report error on it.

BTW, the origin of the `config.node` solution for `fs and child_process not found` error comes from [here](https://github.com/nuxt-community/dotenv-module/issues/11#issuecomment-376780588). You can add the link to the code comment that @greatislander suggested to add, which makes sense.

assets/css/layout/_layout.scss
TeddTech: @greatislander the left side of the background is uneven otherwise. But I see it is causing other problems. I spent a lot of time on this yesterday. Can we do a one on one? I need help figuring this out.

package.json
TeddTech: `fs` and `child_process` are dependencies that are needed to use jsdom which is being implemented in the shared/SideMenu.js file

package.json
TeddTech: I tried figuring this out by removing this dependency from package.json then rebuilding and see what errors come up. After doing this there were no errors but after pushing those changes the test on the PR failed. I then readded those dependencies.

package.json
TeddTech: I tried figuring this out by removing this dependency from package.json then rebuilding and see what errors come up. After doing this there were no errors but after pushing those changes the test on the PR failed. I then readded those dependencies.

cindyli: Please take a look on test failures reported on PR and figure out why they occur on CI but not on your local computer. It will help us to understand which part is problematic. Thanks.

greatislander:
@cindyli @TeddTech and I took a look at this. We got the deploy issues resolved, but encountered a different issue. Because the `dataFetcher` is called on page loads, we are now loading `jsdom` client-side, which has increased the JavaScript bundle size by 3.4MB. I know the routes are being generated statically now but it still seems that we are using `dataFetcher` to populate the store on the initial page load and then loading page content from the store. Can we use a different technique to write the content to each page during the `generate` routine so that we don't need to populate the store? I know this would mean coming up with a different approach for search but I think that would be better.

Thanks, let's discuss at our meeting tomorrow.

greatislander:
Resolved in #207.

Issue #184, opened 2020-04-08
Add a search box on "Views" page


cindyli:
## Is your feature request related to a problem? Please describe.

"Views" page has a "search posts" box.

## Describe the solution you'd like

Implement the search box according to [the design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=0%3A1).

## Describe alternatives you've considered

N/A.

## Additional context

N/A.

PR #203, opened 2020-04-23
Add search to "Views" page

cindyli:

## Description

Add the search functionality to the "Views" page.

## Steps to test

1. Go to "Views" page
2. A search box shows at the top right corner. Its styling should match [the figma design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=239%3A0)
3. Enter a keyword in the search box, hit enter, the search result shows views post with the matched keyword
4. Search the same keyword using the search box in the header, the search result shows all matches across the site.

**Expected behavior:** 

The search via the search box on the "Views" page only searches thru views posts while the search using the search box on the header searches across the website.

### DIFF
--- a/components/SearchForm.vue
+++ b/components/SearchForm.vue
@@ -1,12 +1,27 @@
 <template>
 	<div class="search-container">
 		<form id="search-form" action="/search">
-			<input
-				name="s"
-				type="search"
-				placeholder="Search..."
-				aria-label="Enter keywords for a site-wide search"
-			>
+			<input :placeholder="placeholder" :aria-label="ariaLabel" name="s" type="search">
+			<input :value="context" type="hidden" name="context">
 		</form>
 	</div>
 </template>
+
+<script>
+export default {
+	props: {
+		placeholder: {
+			type: String,
+			default: "Search..."
+		},
+		ariaLabel: {
+			type: String,
+			default: "Search"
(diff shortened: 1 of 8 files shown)


components/SearchForm.vue
```diff
@@ -1,12 +1,27 @@
...
 	<div class="search-container">
 		<form id="search-form" action="/search">
-			<input
-				name="s"
-				type="search"
-				placeholder="Search..."
-				aria-label="Enter keywords for a site-wide search"
-			>
+			<input :placeholder="placeholder" :aria-label="ariaLabel" name="s" type="search">
+			<input :value="context" type="hidden" name="context">
```
TeddTech, line 5: Is this correct how there are two inputs elements here?

cindyli: Having two input elements is for both of them to be included in the URL at the form submission so that the search page knows the scope that the search should be performed. Now the search script handles 2 cases:

1. if the `context` value in the query is empty, search across the site;
2. if the `context === "views"`, only search thru views posts.

cindyli:
The scss work in this pull request will be useful when working on https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/184 for the migrated site using Eleventy.

PR #204, opened 2020-04-24
fix: set different WP domains for different netlify contexts - DO NOT MERGE

cindyli:

## Description

To experiment how to set different WP domains for different netlify contexts.


netlify.toml
```diff
@@ -0,0 +1,5 @@
+[context.deploy-preview.environment]
+	WP_DOMAIN = "https://wecount-dev.inclusivedesign.ca"
+
+[context.branch-deploy.environment]
+	WP_DOMAIN = "https://wecount-cms.inclusivedesign.ca"
```
greatislander, line 5: Brilliant, @cindyli!

cindyli: Unfortunately, it only works at the build time running `npm run generate` but not at the runtime when accessing webpages. I found `process.env.CONTEXT` is not picked up at the runtime too. It always returns `undefined` at browsing webpages.

Still in the process of figuring out how to pass environment variable values from the build time to runtime.

greatislander: I think that is what this is supposed to do: https://nuxtjs.org/api/configuration-env/

cindyli: Works perfectly. Thanks, @greatislander!

.eslintrc.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
 		"quotes": [2, "double", { "avoidEscape": true }],
 		"import/no-named-as-default": 0,
 		"semi": [2, "always"],
-		"vue/html-indent": [1, "tab"]
+		"vue/html-indent": [1, "tab"],
+		"no-console": "off"
```
greatislander, line 23: I'd rather we just add an ignore to the relevant lines in the source file than ignoring this altogether. See: https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

cindyli: "no-console" is turned off so I can debug the deploy preview. It won't be included in the pull request that will be merged into the dev and master.

Will close this pull request soon once the proper pull request is ready.

cindyli:
Closed in favor of https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/205.

PR #206, opened 2020-04-27
Demo of Fully Static Nuxt Site

TeddTech:

## Description

Demo of fully static nuxt site

## Steps to test

1. run `npm run generate`
2. run `npx serve dist`
3. open up network tab in inspect and click around
4. monitor network tab

**Expected behavior:** 

Only making api calls during search


nuxt.config.js
```diff
@@ -44,6 +46,29 @@ export default {
 			{ rel: "icon", type: "image/png", href: "/favicon.png" }
 		]
 	},
+	render: { resourceHints: false },
+	hooks: {
+		// remove uneeded js scripts
+		"generate:page": (page) => {
+			const doc = cheerio.load(page.html);
+			doc("body script[src='/_nuxt/aac6f7dcb0388507b165.js']").remove();
```
cindyli, line 54: Removing javascript on static pages using hardcoded nuxt generated names is not reliable as the script names could change from build to build. I did get a different set of names on my computer.

I wonder if you really need this `hooks` block to remove javascript from static pages? When testing this pull request, I noticed the fetch of the data comes from the local payloads that are saved in the `dist` directory with your addition of `payload` field in the nuxt.config.js `generate` block, which should be enough to deal with the remote fetching problem.

pages/_slug/index.vue
```diff
@@ -37,8 +26,12 @@ export default {
...
 		};
 	},
-	fetch ({ store }) {
-		return store.dispatch("fetchSitePages");
+	asyncData ({ $axios, $payloadURL, route, store }) {
+		if (process.static && process.client && $payloadURL) {
+			return $axios.$get($payloadURL(route));
+		}
+
+		return store.state.sitePages.find(oneSitePage => oneSitePage.slug === route.params.slug);
```
cindyli, line 34: Shall  `await store.dispatch("fetchSitePages");` be added before this line?

pages/index.vue
```diff
@@ -41,8 +35,12 @@ export default {
...
 		};
 	},
-	fetch ({ store }) {
-		return store.dispatch("fetchSitePages");
+	asyncData ({ $axios, $payloadURL, route, store }) {
+		if (process.static && process.client && $payloadURL) {
+			return $axios.$get($payloadURL(route));
+		}
+
+		return store.state.sitePages.find(oneSitePage => oneSitePage.slug === "home");
```
cindyli, line 43: Same as above: add `await store.dispatch("fetchSitePages");` before this line?

pages/_slug/index.vue
TeddTech: No need because it is called in `actions.js` within `nuxtServerInit`

pages/index.vue
TeddTech: No need because it is called in `actions.js` within `nuxtServerInit`

nuxt.config.js
TeddTech: Yes, you are absolutely right. It is enough to deal with the remote fetching problem.

I included the removal of js scripts to show what is possible and different directions we can go. I will expand on this further in the check-in today.

pages/index.vue
cindyli: Makes sense. Thanks.

Issue #214, opened 2020-04-28
Search locally


greatislander:
A full content index of news, views, and pages is available here: https://deploy-preview-207--wecount.netlify.app/index.json (relative link: `/index.json`). We should update the search code in https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/eleventy/src/search.njk to fetch the JSON data and search through it instead, returning the matching items.

We should also display the items using the standard grid layout, the markup for which can be found here:

https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/eleventy/src/_includes/components/grid-item.njk

PR #223, opened 2020-05-01
Search locally

cindyli:
* [X] This pull request has been tested by running `npm run build` and reviewing generated site

## Description

Resolve https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/214.

Perform the search by fetch pre-generated data from `{origin-url}/index.json` rather than using the WordPress search API. It removes the dependency on the backend WordPress completely once the website is built.

One issue I noticed is when a change to data-fetcher requires `index.json` to re-populate but `index.json` already exists, `index.json` is not automatically re-populated with the new data. This issue can be fixed by manually removing `index.json` before rebuild. I'm not sure if the problem will happen with Netlify. Let's keep an eye on it.

## Steps to test

1. Perform a search;
2. Watch the "network" tab on the developer tool;
3. There shouldn't have any call to the WordPress API;
4. Search results should be displayed in the proper grid layout.


greatislander:
> One issue I noticed is when a change to data-fetcher requires `index.json` to re-populate but `index.json` already exists, `index.json` is not automatically re-populated with the new data. This issue can be fixed by manually removing `index.json` before rebuild. I'm not sure if the problem will happen with Netlify. Let's keep an eye on it.

This problem won't happen with Netlify, because every deployment is a fresh build.

TeddTech:
Noticed when Search return more than 10 results, the pages are not paginated.

To recreate this issue query `" "` in search

cindyli:
Good catches, @TeddTech. I will address them when working on https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/225. Thanks.

PR #261, opened 2020-06-02
fix: set the minimum value for UIO line space to 1 (resolves #253)

### LINKED ISSUE #253: Title of posts are cut off on News and Views page
## Describe the bug

Title of posts are cut off on News and Views page

## To Reproduce

Configure UIO to the following 
* Line spacing: 0.7
* Text Style: Comic Sans, Open Dyslexic
* Contrast: any but default
* Enhance Inputs: ON

1. Go to News or Views page
2. Observe post titles

## Expected behavior

Post title Should not be cut off

##### With contrast other than default
Comic Sans
Screen Shot 2020-05-29 at 3 09 31 PM
Open Dyslexic
Screen Shot 2020-05-29 at 3 09 20 PM

##### With default contrast
Screen Shot 2020-05-29 at 3 51 48 PM

cindyli:

## Description

To resolve #253.

Set the minimum value for the UIO line space to 1.

## Steps to test

1. Go to any webpage and open UIO;
2. The value in the UIO line space input field is 1;
3. Cannot use "-" button to the left of the input field to decrease the value;
4. Select a high or low contrast theme;
5. Go to the "news" page

**Expected behavior:** 

News titles are shown properly without any cut-off on texts.

### DIFF
--- a/src/js/uio.js
+++ b/src/js/uio.js
@@ -1,0 +1,49 @@
+/* global $, fluid */
+
+$(document).ready(function () {
+	fluid.defaults("fluid.prefs.schemas.lineSpace", {
+		gradeNames: ["fluid.prefs.schemas"],
+		schema: {
+			"fluid.prefs.lineSpace": {
+				"type": "number",
+				"default": 1,
+				"minimum": 1,
+				"maximum": 2,
+				"multipleOf": 0.1
+			}
+		}
+	});
+
+	fluid.uiOptions(".flc-prefsEditor-separatedPanel", {
+		preferences: [
+			"fluid.prefs.lineSpace",
+			"fluid.prefs.textFont",
+			"fluid.prefs.contrast",
+			"fluid.prefs.tableOfContents",
+			"fluid.prefs.enhanceInputs",
+			"fluid.prefs.syllabification"
+		],
+		auxiliarySchema: {
+			terms: {
+				"templatePrefix": "/lib/infusion/src/framework/preferences/html",
+				"messagePrefix": "/lib/infusion/src/framework/preferences/messages"
+			},
+			"fluid.prefs.tableOfContents": {
+				enactor: {
+					"tocTemplate": "/lib/infusion/src/components/tableOfContents/html/TableOfContents.html",
+					"tocMessage": "/lib/infusion/src/framework/preferences/messages/tableOfContents-enactor.json"
+				}
+			},
+			"fluid.prefs.syllabification": {
+				enactor: {
+					terms: {
+						patternPrefix: "/lib/infusion/src/lib/hypher/patterns"
+					}
+				}
+			}
+		},
+		prefsEditorLoader: {
+			lazyLoad: true
+		}
+	});
+});
--- a/src/js/uio.js
+++ b/src/js/uio.js
@@ -1,19 +1,6 @@
 /* global $, fluid */
 
 $(document).ready(function () {
-	fluid.defaults("fluid.prefs.schemas.lineSpace", {
-		gradeNames: ["fluid.prefs.schemas"],
-		schema: {
-			"fluid.prefs.lineSpace": {
-				"type": "number",
-				"default": 1,
-				"minimum": 1,
-				"maximum": 2,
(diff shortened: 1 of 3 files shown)


src/js/uio.js
```diff
@@ -0,0 +1,49 @@
...
+		schema: {
+			"fluid.prefs.lineSpace": {
+				"type": "number",
+				"default": 1,
+				"minimum": 1,
+				"maximum": 2,
+				"multipleOf": 0.1
+			}
+		}
+	});
```
jobara, line 15: @cindyli this is interesting. You redefined the `fluid.prefs.schemas.lineSpace` schema grade from Infusion. I think we might need to ask @amb26 about this approach.

amb26: This approach is dubious and should be replaced with an options distribution, grade linkage, or other such thing - otherwise, please issue a pull against Infusion!

jobara: @amb26, I'm not completely sure, but I suspect that the issue may revolve around the difficulty of specifying an alternative range for the preference in a particular instance. These values default to what comes from the primary schema, but may not be reflective of what is desired in a particular instance of a UIO/Prefs Editor. @cindyli could you please confirm?

cindyli: @amb26, as @jobara said, it's not easy to use `distributeOptions` or grade linkage to override values that come from the primary schema.

The primary builder uses [a function](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/PrimaryBuilder.js#L54) to look up all primary schema grades registered with the prefs framework using the base grade `fluid.prefs.schemas`. The returned value of this function is then manipulated by [`{that}.buildPrimary`](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/PrimaryBuilder.js#L48) that returns an array of a set of different grade names that are then attached onto the component tree.

This means, even though the value i wanna change is from a primary schema grade named `fluid.prefs.schemas.lineSpace`, this grade is not on the component tree after going thru the primary builder process.

I think this leaves me with 2 options:
1. The easy way is to override `fluid.prefs.schemas.lineSpace` directly so the primary builder will read my version;
2. The difficult way is to dig into the code to figure out how `fluid.prefs.schemas.lineSpace`  gets transformed and where the value I wanna change is ended up on the component tree. Then use `distributeOptions` to override this value.

I lean towards using the easy way for now and file a JIRA against the prefs framework for an improvement.

@amb26, @jobara, what do you think?

cindyli:
@amb26, @jobara helped to figure out a better way to override default values coming from the primary schema. I've also created [a JIRA](https://issues.fluidproject.org/browse/FLUID-6514) for the future improvement of the overriding process.

This pull request is ready for another review. Thanks.

amb26:
Yes, reasonable idea from @jobara - if it is really the case that we can't override the primary schema values via Infusion's reuse schemes this would certainly be worth a JIRA if not an entry in https://wiki.fluidproject.org/display/fluid/Bestiary+of+Reuse+Failures . @cindyli - when you say "fluid.prefs.schemas.lineSpace, this grade is not on the component tree after going thru the primary builder process" - isn't it the case that this grade ends up on fluid.prefs.primaryBuilder itself and so could be targetted there?

cindyli:
> @cindyli - when you say "fluid.prefs.schemas.lineSpace, this grade is not on the component tree after going thru the primary builder process" - isn't it the case that this grade ends up on fluid.prefs.primaryBuilder itself and so could be targetted there?

@amb26, yes, you are right. The grade `fluid.prefs.schemas.lineSpace` ends up as a composing grade of the primary builder, which is a composing grade of `fluid.uiOptions`. The difference is, after going thru [the special `fluid.prefs.schemas` merge policy](https://github.com/amb26/infusion/blob/FLUID-6145/src/framework/preferences/js/PrimaryBuilder.js#L137-L139), values defined in `fluid.prefs.schemas.lineSpace` are moved into a slightly different path named `schema.properties` rather than its original `schema` path. This new path is what @jobara helped to discover and in this pull request now.

PR #285, opened 2020-06-12
fix: non-views posts on search results don't keep preview image blocks

cindyli:

## Description

Non-views posts on search results don't keep preview image blocks

## Steps to test

1. Search for the keyword "ai";
2. Non-views posts on the search result should not keep space for preview images.

**Expected behavior:** 

See above.

### DIFF
--- a/src/scss/abstracts/_mixins.scss
+++ b/src/scss/abstracts/_mixins.scss
@@ -19,6 +19,10 @@
 		right: 0;
 		top: 0;
 	}
+
+	&:empty::before {
+		padding-top: 0;
+	}
 }
 
 // To adapt to UIO contrast themes


TeddTech:
@cindyli lgtm but I am thinking for this issue should we make it so there is always a default image, even in the case theres no given image available for an item

cindyli:
> @cindyli lgtm but I am thinking for this issue should we make it so there is always a default image, even in the case theres no given image available for an item

A good question. I feel having default images for all types of posts might be confusing, for example, news posts don't have preview images by default. Using placeholder preview images for them might give a misleading connection with views posts. 

Let's check with designers at Monday's check in.

TeddTech:
> > @cindyli lgtm but I am thinking for this issue should we make it so there is always a default image, even in the case theres no given image available for an item
> 
> A good question. I feel having default images for all types of posts might be confusing, for example, news posts don't have preview images by default. Using placeholder preview images for them might give a misleading connection with views posts.
> 
> Let's check with designers at Monday's check in.

Good point, I didn't think about that. You're absolutely right.

PR #307, opened 2020-06-22
feat: add dockerfile

gtirloni:

## Description

Add Dockerfile

## Steps to test

1. docker build -t wecount .
2. docker run --rm -ti -p 8000:80 wecount
3. open http://localhost:8000

## Additional information

I know we're using Netlify for deployments but I'd like to have this fallback option readily available so the website can be deployed in our infrastructure.


cindyli:
Thanks for the PR. It's a useful fallback. It would be helpful to add steps of using docker into the README.

I ran into a couple of issues:

1. Running `docker build -t wecount .` throws this error:
```
$ docker build -t wecount .
Sending build context to Docker daemon  1.147MB
Step 1/8 : FROM node:12.18.0-alpine AS builder
Get https://registry-1.docker.io/v2/library/node/manifests/12.18.0-alpine: unauthorized: incorrect username or password
```
Your suggestion of manually pulling the docker image using `docker pull node:12.18.0-alpine` helps. We probably should add this issue and solution into the README too.

2. Some pages are somehow inaccessible. Steps to reproduce:
* Open http://localhost:8000 in a browser;
* Go to "Views" page and click on a post for example "post 56";
* It opens a URL `http://localhost/views/post-56/` which is incorrect. The correct domain should be `http://localhost:8000`.

The interesting thing is when hovering on a views post, the link does point to the correct URL but opening it directs to the wrong one. This problem is not reproducible when the website is started using `npm run start` or `npm run build`.

gtirloni:
> 1. Running `docker build -t wecount .` throws this error:

I think that was a temporary error.

> It opens a URL http://localhost/views/post-56/ which is incorrect. The correct domain should be http://localhost:8000.

I can reproduce the issue if I click on the post #56 link, coming from Views. But if I access that URL directly, it works fine. It seems something in the Javascript is changing it?

gtirloni:
@cindyli could you try building the image again? I think I've found the error. See this commit: https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/307/commits/58548a91322ae40663a11410471460e66df3c3ec

cindyli:
Nice catch and fix, @gtirloni. It does fix the issue.

PR #318, opened 2020-07-08
feat: add filter to "views" landing page (resolves #174)

### LINKED ISSUE #174: Add filter to "Views" landing page

[The design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website) has a "filter" section for "Views" landing page. It filters views articles by tag.

Add the filter according to [the design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website).

N/A.

## Additional context

N/A.

cindyli:

## Description

Implement the filter on the "Views" landing page.

## Steps to test

1. Go to "Views";
2. The filter should be displayed on the page under the "search posts" container;
3. Perform filter only, filter then search, search then filter etc.

**Expected behavior:** 

1. The filter action should filter views by tags;
2. The filter with search action should give results that match both the search and filter conditions;
3. Styles should match the figma design.


cindyli:
>     * One thing I found is that the hover styling for the filter arrow looks odd to me. The fill is not covering the full arrow. I am not sure if this is an issue or not. See image below.
> 
Based on the discussion at today's check in meeting, the focus/hover style for the expand button will hide the outer stroke. This has been addressed.

>     * Another thing I discovered was that after clicking the filter arrow the hover styling remains active even when the user points the mouse away. It remains like this until something else is clicked.
>
This is expected because the focus style for the expand button is same as the hover style. Although the mouse has moved away from the button, the focus is still on it as you just clicked on it so the focus style is still applied. The focus style will be removed the you click elsewhere which moves the focus away from the expand button.

>     * When using tab focus on an active "Tab Filter" there is no styling to show it is selected.
> 
>     * There is no Enhance Inputs styling for the Tags Filter
> 
Based on today's discussion, @uttaraghodke will think about styles for above two issues and get back.

>     * I was thinking it would be nice to have the tags in the "Applied Tags" section as links just like how they are in the [Tags section of individual view post pages](https://deploy-preview-318--wecount.netlify.app/views/journeys-through-the-digital-innovation-appendix-of-the-master-innovation-and-development-plan#tags)

Based on today's discussion, we'll check with Vera at Monday's meeting regarding whether tags in "applied tags" section should be links directing to the tags page.

src/scss/components/_filter.scss
```diff
@@ -0,0 +1,196 @@
...
+	.reset-button {
+		&:focus,
+		&:hover {
+			svg {
+				background-color: $gold-light;
+			}
+		}
+	}
+
+	.filter-expand-button:focus,
```
TeddTech, line 155: For the Tags Filters svg arrow Icon. I believe the solution @uttaraghodke decided on at todays meeting was to have it filled when expanded and not filled other wise. A styling that is uses these selectors `button[aria-expanded="true"]` or `.filter-body[style="display: block;"]` would better accomplish this. 

Also using `stroke-width: 0.25;` inside the selector instead of changing the colour of the stroke would prevent the icon form looking smaller while filled and Tag Filters is expanded.

src/scss/abstracts/_mixins.scss
```diff
@@ -25,35 +25,111 @@
...
+			--expand-button-color: #{$bColor};
+			--expand-button-stroke-color: #{$fColor};
+
+			background-color: #{$bColor} !important;
+
+			svg {
+				background-color: #{$bColor} !important;
+			}
+
+			&:focus,
```
TeddTech, line 110: Same as other comment but for uio contrast themes

src/scss/components/_filter.scss
cindyli: > For the Tags Filters svg arrow Icon. I believe the solution @uttaraghodke decided on at todays meeting was to have it filled when expanded and not filled other wise. A styling that is uses these selectors `button[aria-expanded="true"]` or `.filter-body[style="display: block;"]` would better accomplish this.
> 
@uttaraghodke: could you clarify the filled expand button should be applied when the button is hovered/focused, or when the filter is expanded?

> Also using `stroke-width: 0.25;` inside the selector instead of changing the colour of the stroke would prevent the icon form looking smaller while filled and Tag Filters is expanded.
>
@TeddTech: do you mean `0.25rem`? This is an equivalent of `4px` which is already [the default stroke width](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/feat/add-filter-to-views/src/scss/components/_filter.scss#L20) for the expand button. It is the cause of the uneven button top you mentioned in the previous comment. I believe our decision from yesterday was to remove the stroke width when the button is filled to fix the uneven button top.

TeddTech: No `stroke-width: 0.25;` is not the same as `stroke-width: 0.25rem;`. By reducing `stroke-width` it can allow the fill to appear bigger.

cindyli:
@TeddTech, this pull request is ready for another review.

src/js/views-dynamic-handler.js
```diff
@@ -1,46 +1,131 @@
 // For search functionality on the header.
 
-/* global Vue, axios, search, createPagination */
+/* global Vue, axios, search, createPagination, processDisplayResults, filter, $ */
 
+const pageSize = 10;
 const params = new URLSearchParams(window.location.search);
-let searchQuery = params.get("s");
-searchQuery = searchQuery ? searchQuery.toLowerCase() : undefined;
+let searchTerm = params.get("s") ? params.get("s").trim() : undefined;
```
TeddTech, line 7: If Tag Filter is used and returns greater than 1 page of results then when you navigate to a different page number using pagination the term "undefined" is queried instead of taking you to the next page of results. Example:

1. Go to Views page
2. Use Tag Filter by only checking _test post_ (this gives multiple pages of results)
3. Click Apply filters
4. Scroll down to pagination
5. Use pagination to navigate to another page of results
6. See error (Instead of navigating to the next page of results "undefined" is being queried)

cindyli: Good catch, @TeddTech. This has been addressed and ready for another look. Thanks.

TeddTech: Hi @cindyli, the error still persist.

cindyli: I'm having a weird problem that my new commit to my branch doesn't get shown on the pull request. Looking into it.

cindyli: Thanks for @jobara and @gtirloni's information, there's [a problem with Github Action](https://www.githubstatus.com/) at the moment. Let's see if new changes will be reflected when the problem is resolved.

PR #321, opened 2020-07-10
fix: improved styling and display of content images from wordpress (resolves #137)

### LINKED ISSUE #137: Images: support content images

We can improve our styling and display of content images from WordPress

We should support all content image orientations from WordPress.

Not applicable.

## Additional context

- https://richtabor.com/gutenberg-image-alignments/

TeddTech:

## Description
This PR improves styling and display of content images from WordPress.

## Steps to test

1. Go to deploy preview.
2. Navigate to Views page.
3. Observe the alignment of images in test posts 89-100.
4. Observe the alignment of media in [Journeys through the Digital Innovation Appendix of the Master Innovation and Development Plan](https://deploy-preview-321--wecount.netlify.app/views/journeys-through-the-digital-innovation-appendix-of-the-master-innovation-and-development-plan/) post and [Project We Count evaluates Toronto’s Quayside Master Innovation and Development Plan](https://deploy-preview-321--wecount.netlify.app/views/project-we-count-evaluates-torontos-quayside-master-innovation-and-development-plan/) post.

**Expected behavior:** 

Images should be sized and aligned according to how they were set in WordPress.
## Additional information
N/A

## Related issues
resolves #137


src/scss/layout/_layout.scss
```diff
@@ -57,17 +43,59 @@ article {
 			display: table;
 		}
 	}
-}
 
-/*
- * Default WP Alignment Classes
+	/*
+ * Default WP image Alignment Classes
```
cindyli, line 48: Please fix the comment: styles for WP selectors in this section are no longer default ones from WordPress but customized for the WeCount website.

src/scss/layout/_layout.scss
```diff
@@ -57,17 +43,59 @@ article {
...
+	}
+
+	.wp-block-image figure .alignfull {
+		margin-left: calc(-100vw / 2 + 100% / 2);
+		margin-right: calc(-100vw / 2 + 100% / 2);
+		max-width: 100vw;
+	}
+
+	.wp-block-image figure .alignwide {
+		margin-left: -100px;
```
cindyli, line 91: Use `rem()` instead of `px` whenever possible. The same comment applies to other usage of `px`.

src/scss/layout/_layout.scss
```diff
@@ -57,35 +43,73 @@ article {
...
+	.wp-block-image figure .alignfull {
+		margin-left: calc(-100vw / 2 + 100% / 2);
+		margin-right: calc(-100vw / 2 + 100% / 2);
+		max-width: 100vw;
+	}
+
+	.wp-block-image figure .alignwide {
+		margin-left: rem(-100);
+		margin-right: rem(-100);
+		max-width: rem(1024);
```
cindyli, line 93: I wonder why `max-width` is set to a fix `1024px` value rather than a relative value such as `100%`? Same question for line 97.

TeddTech: Good catch! The value for these lines should be `100%`.

src/scss/layout/_layout.scss
```diff
@@ -57,35 +43,73 @@ article {
 			display: table;
 		}
 	}
-}
 
-/*
- * Default WP Alignment Classes
+	/*
```
cindyli, line 47: It would be nice to use comment styles that have been applied elsewhere here, for example: `/** Customized styles for WP content image selectors **/`.

src/scss/layout/_layout.scss
```diff
@@ -57,35 +43,73 @@ article {
...
+		}
+
+		.alignright {
+			float: right;
+			margin-left: rem(25);
+			margin-top: 0;
+		}
+	}
+
+	.wp-block-image figure .alignfull {
```
cindyli, line 84: If you inspect DOM structures when `alignfull` or `alignwide` are applied, correct selectors for them are `.wp-block-image.alignfull` and `.wp-block-image.alignwide`. With wrong selectors here, the css from line 84-98 has never been applied. These 2 cases are actually using the default `figure` css at line 50. Please investigate and fix. Thanks.

TeddTech: Tried it out and you are right. Looks like there may not be a need for the selectors on line 84-98.

cindyli: I agree.

TeddTech: Okay, I will remove them.

TeddTech:
As discussed in today's we count check-in meeting I will add support for WordPress Media & Text blocks in addition to WordPress Image blocks.

cindyli:
We should add support for "audio" blocks too.

TeddTech:
Ready for review. It may help to refer to [test post 89](https://deploy-preview-321--wecount.netlify.app/views/test-post-89/) and [test post 88](https://deploy-preview-321--wecount.netlify.app/views/test-post-88/)

src/scss/layout/_layout.scss
```diff
@@ -57,35 +43,89 @@ article {
...
-	text-align: center;
+	// WP customized styles for the WeCount website
+
+	// WP styles for vertically aligning text within Media & Text blocks
+	.is-vertically-aligned-center .wp-block-media-text__content {
+		align-items: center;
+		display: flex;
+	}
+
+	.is-vertically-aligned-bottom .wp-block-media-text__content {
```
cindyli, line 55: You've defined the vertical alignment for "center" and "bottom" but not for "top". By looking at the computation of [the default value "auto" for `align-self`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self#Specifications), the result depends on the computed value of the parent container. To avoid the unexpected behavior, it's probably better to explicitly specify `align-self` value for `.is-vertically-aligned-top` too.

cindyli: we should support `.has-large-font-size` selector.

src/scss/layout/_layout.scss
```diff
@@ -107,35 +147,85 @@ figcaption {
...
+				}
+			}
+		}
 	}
+}
 
-	.alignleft {
-		float: left;
-		margin: 0.5em 1em 1em 0;
+@media screen and (min-width: rem(1366)) {
```
cindyli, line 221: What's the reason to define a new breakpoint at 1366px? The css in this section looks like a duplicate of line 204-208.

TeddTech: The reason to define this breakpoint is to adjust the size of aligned left or aligned right embedded media on desktop screens. Lines 210-215 adjust the size of aligned left or aligned right embedded media for iPad Pro screens. Lines 204-208 is the sized of embedded media that is NOT aligned right or aligned left for desktop and iPad Pro screens.

TeddTech: The reason to define this breakpoint is to adjust the size of aligned left or aligned right embedded media on desktop screens. Lines 210-215 adjust the size of aligned left or aligned right embedded media for iPad Pro screens. Lines 204-208 adjust the size of embedded media that is NOT aligned right or aligned left for desktop and iPad Pro screens.

cindyli: Thanks for the explanation. The larger media iframe does look better on larger screen. Can you add a code comment here. Please also include that picking `1366px` is the starting px for ipad pro landscape mode.

PR #358, opened 2020-07-30
Galaxy fold search bar (resolves #355)

### LINKED ISSUE #355: Search bar covers the menu button in the Galaxy fold screen when enhanced inputs are turned on
## Describe the bug

Search bar covers the menu button in the Galaxy fold screen when enhanced inputs are turned on.

The solution is to use a separate media query to hide text “Menu” for galaxy fold.

## To Reproduce

1. Go to any page and view in the Galaxy fold mobile screen;
2. Open UIO; enable "enhance inputs"; set font style to "Open Dyslexic";
3. Click the search icon on the navigation bar;
4. The search box overlaps with the text "Menu".

## Expected behavior

The overlapping should not happen.

## Screenshots

Screen Shot 2020-07-30 at 5 31 30 PM

TeddTech:

## Description

This PR removes the "Menu" text in the header nav on the Galaxy Fold screen size.

## Steps to test

1. Go to deploy preview
2. Inspect using Galaxy Fold screen
3. Click on the search bar in the header nav

**Expected behavior:** 
Search bar should not overlap menu icon when expanded.
## Additional information
N/A

## Related issues
closes #355


src/_includes/components/header-nav.njk
```diff
@@ -20,7 +20,7 @@
 <div class="nav-smallScreen">
 	<nav class="starter-nav">
 		<button id="menuToggleButton" aria-expanded="false">
-		  <svg><use xlink:href="#menu-icon" /></svg>&nbsp;Menu
+		  <svg><use xlink:href="#menu-icon" /></svg>&nbsp;
```
cindyli, line 23: Removing the text "Menu" will introduce an accessibility issue that this menu icon will not be recongnized as a menu button by ATs. It's probably better to wrap the "menu" text in a div or span and show/hide it when needed.

src/_includes/components/header-nav.njk
```diff
@@ -20,7 +20,7 @@
 <div class="nav-smallScreen">
 	<nav class="starter-nav">
 		<button id="menuToggleButton" aria-expanded="false">
-		  <svg><use xlink:href="#menu-icon" /></svg>&nbsp;Menu
+		  <svg><use xlink:href="#menu-icon" /></svg>&nbsp;<span>Menu</span>
```
cindyli, line 23: I think I missed one point in my previous comment. When the menu span is hidden on the Galaxy Fold, the same accessibility issue will occur that ATs won't be able to access this "Menu" text.

A possible fix is to add a `screen-reader-only` span that will be shown on Galaxy Fold and hidden on other screen sizes. Hiding on other screen sizes will prevent the "menu" text being read/accessed twice from 2 different spans.

TeddTech: Ready for another review. I am not sure if I did it correctly but I believe this new push is what you are talking about.

cindyli: Awesome fix. Thanks, @TeddTech.

Issue #359, opened 2020-07-31
h1 break error


Raunaksingh100:
## Is your feature request related to a problem? Please describe.

H1 error breaking when we use text-style with enable Syllabus ....

## Describe the solution you'd like

There is no need of adding ""hyphen:auto ""in this i added a  customized font-size (by their I'd),,,  
After  this I  won't get any bug , in any type of the screen 

**You can check the error **

1. Here the link direct link -

https://wecount.inclusivedesign.ca/views/journeys-through-the-digital-innovation-appendix-of-the-master-innovation-and-development-plan/

2. select the **OPEN DYSLEXIC** text-style with enable the syallbus feature you get this bug

### screenshot 

![error](https://user-images.githubusercontent.com/65535360/89003744-3337dc80-d31e-11ea-9f6c-f43ede19cf47.PNG)

## Additional context
I am working on this issue

### PLZ VIEW MY PR FOR BETTER UNDERSTANDING this works in all devices plz accept this idea its better then using hypens

cindyli:
This ticket is a subset of #256. Closed.

PR #360, opened 2020-07-31
Your raunak fixed h1

Raunaksingh100:

## Description

this pr removed the bug in the h1 completely 
1. you can use now anytype of textstyle with enable the syallbus you won't get any line break,,,

2. in tested this in all types of screens size like -galaxy old , iphones 
you can test in any type of phone you wont get any type of bug

3. this bug is fixed  after changing the font-size and after doing this the webite won't affect infact it become more good and readable

## Steps to test

1. go to deploy preview 
2. you can try with any type of text style with syallbus feature on you won't get any bug in any screen
3. 

### screenshot before
error

### after
fixed

## Related issues
#359

### DIFF
--- a/src/scss/base/_base.scss
+++ b/src/scss/base/_base.scss
@@ -125,3 +125,9 @@ button::before:not(.fl-prefsEditor-buttons) button {
 		}
 	}
 }
+
+@media screen and (min-width: 320px) and (max-width: 568px) {
+	.title {
+		font-size: 2rem;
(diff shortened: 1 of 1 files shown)


Raunaksingh100:
@cindyli plz accept this idea Because  this idea fixed this error in all type of devices you can also check this bug in different devices you won't get any bug
if you get any bug then you reject my pr............

PR #378, opened 2020-08-26
feat: create the workshop comments page (resolves #350)

### LINKED ISSUE #350: Create the new workshop comments page

Clicking a workshop on the "Initiatives" page redirects to a new workshop page that shows the detail of this workshop and a comment area where users can post new comments and view existing comments.

The design for this page can be found on [the Figma for the "Initiatives" page](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=1881%3A3695).

## Some implementation points
* Verify name and comment values at the client side: Name and Comment values are required
* When the "post a comment" button is clicked, send a POST request to a Netlify function endpoint at the server side with a payload of comment contents to save
    * The server side Netlify function will be implemented using #370
* Page style

TeddTech:

## Description

Created workshop page that shows the details of a given workshop and a comment area where users can post new comments and view existing comments.

## Steps to test

1. Go to deploy preview
2. Navigate to initiatives page
3. Select a workshop by clicking on its image
4. Submit a comment with missing fields by clicking on the "post a comment" button to test "Required" message
5. Submit a comment with no fields missing by entering a name and comment then clicking on the "post a comment button"

**Expected behavior:** 

* When any of the fields are missing before clicking on the "post a comment" button a "Required" message should be displayed beside the field label
* When no fields are missing before clicking on the "post a comment" button a confirmation message and the new comment should be displayed.

## Additional information

Aria labels still need to be added to the comments page.

Netlify functions can only run on Netlify servers so this PR has multiple pushes for testing purposes denoted by commit message format `fix: api data test #` where # is the test number.

## Related issues

closes #350


src/js/workshop-comments.js
```diff
@@ -0,0 +1,81 @@
...
+		// Bind the FormData object and the form element
+		const FD = new FormData( form );
+
+		// Define what happens on successful data submission
+		XHR.addEventListener( "load", function() {
+			setSubmittedComment();
+		});
+
+		// Define what happens in case of error
+		XHR.addEventListener( "error", function() {
```
cindyli, line 22: If the `POST` request to server fails, it's probably better to display the error on the UI so users know what's going on. The displayed message should contain the error returned from the server.

@uttaraghodke, the last page for "Initiatives" shows the successful message when the comment is successfully posted. Could you make a design of how the error should be displayed when the comment is NOT saved successfully. Thanks.

src/initiatives.njk
```diff
@@ -1,6 +1,9 @@
 ---
 title: Initiatives
-permalink: /initiatives/
+permalink: /initiatives/{% if pagination.pageNumber > 0 %}page/{{ pagination.pageNumber + 1 }}/{% endif %}
```
cindyli, line 3: Without having the pagination bar on the page for users to switch between pages, adding pagination to `permalink` doesn't make sense.

src/scss/components/_initiatives.scss
```diff
@@ -1,16 +1,218 @@
+.workshop-image {
+	max-width: 100%;
+}
+
+.comments {
+	margin-top: rem(50);
+
+	p:first-child {
```
cindyli, line 8: This selector to set the first comment to grey light is not useful as line 17-23 also sets the same for the submitted comment.

src/scss/components/_initiatives.scss
```diff
@@ -1,16 +1,218 @@
...
+		border-bottom: rem(5) solid $gold;
+		font-size: rem(24);
+		line-height: rem(60);
+	}
+}
+
+.submitted-comment-message {
+	display: none;
+	margin: auto auto;
```
cindyli, line 49: Using one `auto` is enough.

src/scss/components/_initiatives.scss
```diff
@@ -1,16 +1,218 @@
...
+#comment-form {
+	margin-bottom: rem(150);
+
+	.asterisk-def {
+		color: $grey-light;
+		display: inline;
+		font-size: rem(20);
+	}
+
+	.form-control {
```
cindyli, line 74: This selector seems attached to form fields not form controls. Please rename. Thanks.

src/scss/components/_initiatives.scss
```diff
@@ -1,16 +1,218 @@
...
+		color: transparent;
+		text-shadow: 0 0 0 $green-dark;
+	}
+}
+
 .initiatives {
 	.workshop {
 		padding: 0;
 
+		.description p:first-child {
```
cindyli, line 183: What is this for?

cindyli:
When focusing on "name" and "comment" fields, the cursor focus is too close to the field border. On firefox, the cursor focus in the comment field even goes out of the rounded border.

cindyli:
Set the focus in the "name" or "comment" fields and type. The font sizes are much smaller than the base font size. It's probably better to use the same font size comment fields. Moreover with firefox, the font size in "name" is small than that in "comment" field.

cindyli:
The time taken from clicking the "post a comment" button until the thanks prompt shows up takes a few seconds. This delay may confuse users and make them clicking more on the post button. We should talk to designers to find out if it's necessary to either disable the button during this period or show a progress icon or do both.

cc @uttaraghodke, @lliskovoi

cindyli:
After user entered comment is successfully submitted, user entered texts remain in text fields. See the screenshot below. Shall fields be cleared?

Screen Shot 2020-09-01 at 10 38 52 AM

cc @uttaraghodke, @lliskovoi

src/initiatives.njk
TeddTech: Okay, this line change should happen in a separate PR where pagination is added to the initiatives page.

src/workshop-comments.njk
```diff
@@ -0,0 +1,54 @@
...
+
+			<div class="form-control">
+				<span>*</span><label for="comment"><h4>Post a comment or question</h4></label><span class="required">Required</span>
+				<textarea id="comment" type="text" name="comment" :value="commentTerm" placeholder="" aria-label="Enter your comment"></textarea>
+				<input type="hidden" name="workshopId" id="workshopId" value="{{ workshopItem.id }}">
+				<button type="submit">Post a comment</button>
+			</div>
+		</form>
+
+		<div class="submitted-comment-message">
```
cindyli, line 40: Status areas announcing successful or failed comment submissions should be marked as [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so that ATs could notify users what happens.

Please test with screen readers such as VoiceOver on Safari to make sure the submission result is announced after clicking "post a comment" button. Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,81 @@
...
+
+	const submittedCommentDate = submittedComment.querySelector(".comment-date");
+	const submittedCommentName = submittedComment.querySelector(".comment-name");
+	const submittedCommentComment = submittedComment.querySelector(".comment-comment");
+
+	const dateObject = new Date();
+	const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
+	submittedCommentDate.innerText = `${months[dateObject.getMonth()]} ${dateObject.getDate()}, ${dateObject.getFullYear()}`;
+
+	submittedCommentName.innerText = name.value.trim();
```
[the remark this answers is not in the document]
TeddTech: @cindyli they come from the [3rd and 4th line respectively](https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/5a3e340095508029d0b4f04291d7e0fa70d98322/src/js/workshop-comments.js#L3-L4)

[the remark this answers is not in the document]
cindyli: I see. Thanks for pointing out.

src/initiatives.njk
TeddTech: This will be addressed in this [issue](https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/383)

src/workshop-comments.njk
```diff
@@ -0,0 +1,58 @@
...
+
+			<div class="form-field">
+				<span>*</span><label for="comment"><h4>Post a comment or question</h4></label><span class="required">Required</span>
+				<textarea id="comment" type="text" name="comment" :value="commentTerm" placeholder="" aria-label="Enter your comment"></textarea>
+				<input type="hidden" name="workshopId" id="workshopId" value="{{ workshopItem.id }}">
+				<button id="post-comment" type="submit">Post a comment</button>
+			</div>
+		</form>
+
+		<div class="submitted-comment-message post-success-message" aria-live="assertive">
```
cindyli, line 40: While looking up for ARIA live region usage for status, I came across these ARIA roles explained by WCAG specs:

1. [Using role=status to present status messages](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22.html)
2. [Using ARIA role=alert or Live Regions to Identify Errors](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19.html)

These articles talk about live region attributes that are implied in these roles.

It feels to me the success message is more like a status message that could use `role=status` while the error message could `role=alert`. What do you think?

cindyli:
1. Use dev tool to set the network to offline, submit a comment;
2. The error message shows up;
3. Set the network back to online, submit a comment;
4. A problem occurs that both success and error messages are showing on the page. See the screenshot below.

The expected behaviour is only the latest status or error message shows.

This problem is related to [this comment above](https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/378#discussion_r485099964). Hiding both message divs at the submission should fix the problem.

Screen Shot 2020-09-08 at 9 46 14 PM

cindyli:
1. Open developer tool, go to "console";
2. Enter a comment and click "post a comment" button;
3. The "thank you" message appears indicating the comment is saved successfully;
4. However, the console shows an error that the post request failed with "400: Bad request";
5. Check airtable, the comment is not saved.

Screen Shot 2020-09-08 at 9 55 17 PM

It indicates two things: 
1. It's better to check the returned status code. Only show the success message when the status code is 200. Otherwise, show error.
2. Figure out why the comment is not saved.

TeddTech:
@cindyli ready for review

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
+const commentForm = document.getElementById("comment-form");
+
+const nameDOMElement = document.getElementById("name");
```
cindyli, line 3: Line 3 & 4 are duplicates of line 9 & 11. It would be better to remove these lines of global scoped variables but rather using the line 9 & 11 that are functional scoped.

As to the use of these variables in `setSubmittedComment()` function, improve this function by accepting two parameters of `name` and `comment` values generated at line 10 & 12 so that `setSubmittedComment()` function doesn't need to re-calculate these values.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+const commentDOMElement = document.getElementById("comment");
+
+commentForm.addEventListener("submit", (event) => {
+	event.preventDefault();
+
+	const nameDOMElement = document.getElementById("name");
```
cindyli, line 9: Sorry that in our pair programming it was me who started to use the all-capitalized "DOM" as a part of the variable name. After thinking more about it, I felt it might be more conventional to have the variable named "nameDomElement". Could you please rename:

1. nameDOMElement -> nameDomElement
2. commentDOMElement -> commentDomElement

Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	const name = nameDOMElement.value.trim();
+	const commentDOMElement = document.getElementById("comment");
+	const comment = commentDOMElement.value.trim();
+
+	const submittedCommentMessage = document.querySelector(".post-failure-message");
```
cindyli, line 14: It might be better to rename the variable "submittedCommentMessage" to "submitFailureMessage".

cindyli: Please rename the DOM selector for the failure message to `.submit-failure-message` for consistence.

Please Rename the corresponding variable and DOM selector for the success message in the way. Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+
+	disableFields(true);
+
+	// Hide required message
+	// Note: It's necessary to hide "required" indicators programmatically even though they have been hidden initially by CSS
+	// in the case the same user submits more comments after submitting an initial comment
+	document.querySelectorAll(".required").forEach(element => {
```
cindyli, line 21: Selector queries at line 21 and 27 can be combined into:

```
document.querySelectorAll(".required, .save-response")...
```

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	});
+	// Hide comment submission message
+	// Note: It's necessary to hide succesful/failed submitted comment message indicators programmatically even though they have been hidden initially by CSS
+	// in the case the same user submits more comments after submitting an initial comment
+	document.querySelectorAll(".save-response").forEach(element => {
+		applyVisibility(element, false);
+	});
+
+	// Verify input Fields
+	if ((name === "") & (comment === "")) {
```
cindyli, line 32: Line 32-38 can be removed as the individual verification for "name" and "comment" will cover it.

The DOM selector ".required" probably can be removed to as individual selectors of ".required-name" and ".required-comment" cover it too.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+
+	if (comment === "") {
+		applyVisibility(document.querySelector(".required-comment"), true, "inline");
+		disableFields(false);
+		return;
+	}
+
+	const XHR = new XMLHttpRequest();
+
+	const commentForm = document.getElementById("comment-form");
```
cindyli, line 54: This line is a duplicate of line 1 and can be removed.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	// Bind the FormData object and the form element
+	const commentFormData = new FormData( commentForm );
+
+	// submit and add comment to top of comment section of page on successful data submission
+	XHR.addEventListener( "load", function() {
+		if (XHR.status === 200) {
+			setSubmittedComment();
```
cindyli, line 61: It's probably better to rename "setSubmittedComment" to "showSubmittedComment".

cindyli: Please move line 121 that shows the success message out of `setSubmittedComment()` function to right calling this function so the function doesn't perform more that what it's supposed to do and the logic here is cleaner.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+		// Show failed comment submission message
+		submittedCommentMessage.setAttribute("style", "display: block;");
+		disableFields(false);
+	} );
+
+	// Set up our request
+	XHR.open( "POST", window.location.origin + "/api/comments" );
+
+	// The data sent is what the user provided in the form
+	const jsonFormData = JSON.stringify(Object.fromEntries(commentFormData));
```
cindyli, line 82: Did you figure out why JSON data wasn't constructed correctly last week?

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	XHR.setRequestHeader("Content-Type", "application/json");
+	XHR.send(jsonFormData);
+});
+
+/*
+ * Hide/show required and success messages
+ * @param {DOMElement} DOMselector - The DOM Element to parse.
+ * @param {Boolean} toDisplay - value is true to show and false to hide
+ * @param {String}[Optional] displayType - CSS display value to apply
```
cindyli, line 91: Improve the comment that `displayType` only applies when `toDisplay` is set to `true`.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+ */
+function applyVisibility(DOMselector, toShow, displayType = "none") {
```
cindyli, line 93: It doesn't make sense to set the default `displayType` to `none` as the `displayType` will be applied when showing the DOM content. It would make more sense to set the default value to "block".

Rename "DOMselector" -> "domSelector".

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	toShow ? DOMselector.setAttribute("style", `display: ${displayType};`) : DOMselector.setAttribute("style", "display: none;");
```
cindyli, line 94: This line can be optimized to:

```
const displayType = toShow ? displayType : "none";
domSelector.setAttribute("style", "display:  ${displayType};");
```

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/*
+ * Add new comment to the given workshop's page temporarily.
```
cindyli, line 98: "Add" -> "Show" or "Display": Probably "Display the new comment on the current page."?

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	newCommentDiv.innerHTML = "<article class='comment submitted-comment'><p><span class='comment-name'></span> | <span class='comment-date'></span></p><p class='comment-text'></p></article>";
+	const commentsDiv = document.getElementsByClassName("comments")[0];
+	commentsDiv.prepend(newCommentDiv.firstChild);
+
+	const submittedCommentDiv = commentsDiv.firstChild;
+	const submittedDate = submittedCommentDiv.querySelector(".comment-date");
+	const submittedName = submittedCommentDiv.querySelector(".comment-name");
+	const submittedComment = submittedCommentDiv.querySelector(".comment-text");
+
+	const dateObject = new Date();
```
cindyli, line 112: line 112-116 is a copy of `convertDate()` in the utils.js? Please directly call that function rather than a copy/paste. Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/*
+ * Disable the "post a comment" button and text fields when the post request is in the process of sending.
+ * @param {Boolean} sendingStatus - Boolean value indicating whether or not post request is in the send state.
+ */
+function disableFields(sendingStatus) {
+	// Comment form submit button
+	const postCommentButton = document.getElementById("post-comment");
+	sendingStatus ? postCommentButton.setAttribute("style", "background-color: #f3f3f3") : postCommentButton.setAttribute("style", "background-color: #fff");
```
cindyli, line 131: Same as the comment above, line 131 and 141 can be optimized to:

```
const backgroundColor = sendingStatus ? "#f3f3f3" : "#fff";
postCommentButton.setAttribute("style", "background-color: " + backgroundColor);
```

Similar change to line 141.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,144 @@
...
+	postCommentButton.disabled = sendingStatus;
+
+	// Comment form input fields
+	const nameField = document.getElementById("name");
```
cindyli, line 135: line 135-140 can be simplified to:

```
document.querySelectorAll("#name, #comment")...
```

src/workshop-comments.njk
```diff
@@ -0,0 +1,58 @@
...
+
+			<div class="form-field">
+				<span>*</span><label for="comment"><h4>Post a comment or question</h4></label><span class="required required-comment">Required</span>
+				<textarea id="comment" type="text" name="comment" placeholder="" aria-label="Enter your comment"></textarea>
+				<input type="hidden" name="workshopId" id="workshopId" value="{{ workshopItem.id }}">
+				<button id="post-comment" type="submit">Post a comment</button>
+			</div>
+		</form>
+
+		<div class="save-response post-success-message" role="status">
```
cindyli, line 40: The selector ".save-response" can be removed as each DOM already has its own selector identifying success or failure message.

src/workshop-comments.njk
```diff
@@ -0,0 +1,58 @@
...
+		<div class="comments">
+			{% for comment in workshopItem.comments %}
+				{% include 'components/comment.njk' %}
+			{% endfor %}
+		</div>
+	</article>
+{% endblock %}
+
+{% block footerScripts %}
+<script src="/js/workshop-comments.js"></script>
```
cindyli, line 57: Include "utils.js" for using `convertDate()` in it.

cindyli (changes requested):

src/js/workshop-comments.js
TeddTech: This case covers when both the `name` field and `comment` field are missing. Without these lines when both fields are missing it will only show the required message for the first field that is empty then hit `return`. (In this case it will only show the required message for the name field when both fields are empty because comes first in the script)

cindyli: I see your point, @TeddTech. A solution is replace `return` in the following verification with setting an error flag such as `isValueMissing = true`. Check the flag after all verification and return if it's `true`.

src/js/workshop-comments.js
TeddTech: YES! It had nothing to do with `FormData( )` object or the structure of the comment form.

Going from the designers suggestions I disabled the input fields while request are being sent but all fields that are disabled are unable to be sent. This is why only the workshopID filed was being sent. (I never disabled it because it is a hidden input field)

I found a [workaround using the `readonly`](https://stackoverflow.com/questions/7357256/disabled-form-inputs-do-not-appear-in-the-request) attribute that gives a similar functionality as the `disable` attribute. Input fields that are read only are sent during form requests. I had to change the colour (to light grey) during request programatically. See the `disableFields(sendingStatus)` function.

src/js/workshop-comments.js
TeddTech: I am getting the following error when trying to implement this change

```js
const displayType: any
Parsing error: Identifier 'displayType' has already been declaredeslint
```

src/workshop-comments.njk
TeddTech: This selector is still needed for the styling of these elements in `_initiative.scss`

TeddTech: This selector is still needed for the styling of these elements in `_initiatives.scss`

cindyli: The styling in the css can point to individual selectors: replace ".save-response" to ".submit-success-message, .submit-failure-message".

src/js/workshop-comments.js
cindyli: Nice discovery and workaround. Thanks.

src/js/workshop-comments.js
cindyli: Right. "displayType" has already been declared as the input argument. The `const` should be removed in the first line: `displayType = toShow ? displayType : "none";`.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,132 @@
...
+	const name = nameDomElement.value.trim();
+	const commentDomElement = document.getElementById("comment");
+	const comment = commentDomElement.value.trim();
+
+	const submitFailureMessage = document.querySelector(".submit-failure-message");
+
+	// Hide required and comment submission message.
+	// Note: It's necessary to hide "required" and "succesful/failed" submitted comment message indicators programmatically even though they have been hidden initially by CSS
+	// in the case the same user submits more comments after submitting an initial comment.
+	document.querySelectorAll(".required, .submit-success-message, .submit-failure-message").forEach(element => {
```
cindyli, line 18: ".required" selector can be removed as those DOM elements are also be identified as ".required-name" and ".required-comment".

src/js/workshop-comments.js
cindyli: Could you add a code comment to warn the problem caused by disabling input fields and the wordaround of using `readonly`. It would also be helpful to include the url to the workaround you discovered. Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,132 @@
...
+	submittedDate.innerText = convertDate(dateObject);
+	submittedName.innerText = nameField.value.trim();
+	submittedComment.innerText = commentField.value.trim();
+}
+
+/*
+ * Disable the "post a comment" button and text fields when the post request is in the process of sending.
+ * @param {Boolean} sendingStatus - Boolean value indicating whether or not post request is in the send state.
+ */
+function disableFields(sendingStatus) {
```
cindyli, line 118: As mentioned before, please use a variable name that make sense to the functionality of this function rather than the context of using this function because it could be used in other contexts too.

It makes more sense to rename `sendingStatus` to something like `disabled`.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,132 @@
...
+	const backgroundColor = sendingStatus ? "#f3f3f3" : "#fff";
```
cindyli, line 119: Add a code comment about changing background color only applies when not applying UIO themes. When an UIO theme is selected, the background color will remain the same of what's set by the theme.

cindyli (changes requested):

cindyli:
1. Select an UIO theme and add a comment;
2. The new comment is added at the top of the comments area.

Problem: The new comment looks same as old comments.
Expected: The style of the new comment should be able to differentiate it from old comments.

One solution is to apply lower the opacity value of the new comment, such as setting it to 0.7 when a UIO theme is selected.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,132 @@
...
+
+	const XHR = new XMLHttpRequest();
+
+	// Bind the FormData object and the form element.
+	const commentFormData = new FormData( commentForm );
+
+	// Submit and add comment to top of comment section of page on successful data submission.
+	XHR.addEventListener( "load", function() {
+		if (XHR.status === 200) {
+			showSubmittedComment(nameDomElement, commentDomElement);
```
cindyli, line 50: Passing in `name` and `comment` as arguments rather than their DOM elements. The clean-up of these values have already been performed here and don't need to do the same in the `showSubmittedComment()` function.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/*
+ * Disable the "post a comment" button and text fields when the post request is in the process of sending.
+ * @param {Boolean} sendingStatus - Boolean value indicating whether or not post request is in the send state.
+ */
+function disableFields(sendingStatus) {
+	const backgroundColor = sendingStatus ? "#f3f3f3" : "#fff";
+
+	// Comment form submit button.
```
cindyli, line 121: Improve the code comment to be clear about what will happen to the submit button. The same comment applies to line 126.

TeddTech:
@cindyli ready for review

src/js/workshop-comments.js
```diff
@@ -0,0 +1,138 @@
...
+	// Set up our request.
+	XHR.open( "POST", window.location.origin + "/api/comments" );
+
+	// The data sent is what the user provided in the form.
+	const jsonFormData = JSON.stringify(Object.fromEntries(commentFormData));
+	XHR.setRequestHeader("Content-Type", "application/json");
+	XHR.send(jsonFormData);
+});
+
+/*
```
cindyli, line 78: See [JSDoc](https://jsdoc.app/about-getting-started.html): "Each comment must start with a `/**` sequence in order to be recognized by the JSDoc parser."

This change applies to all comments for functions.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {String}[Optional] displayType - CSS display value to apply. Only applicable when toShow is set to true.
+ */
+function applyVisibility(domSelector, toShow, displayType) {
+	displayType = toShow ? displayType : "none";
+	domSelector.setAttribute("style", `display:  ${displayType};`);
+}
+
+/*
+ * Display the new comment on the current page.
+ * @param {DOMElement} nameField - DOM element of name field in comment form.
```
cindyli, line 91: Arguments are no long DOM elements. Please fix the function doc.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {DOMElement} commentField - DOM element of comment field in comment form.
+ */
+function showSubmittedComment(nameField, commentField) {
```
cindyli, line 94: It's probably better to change "nameField" to "name" and "commentField" to "comment" as they are values instead of fields.

cindyli (changes requested):

cindyli:
The "post a comment" button looks broken with Safari.

cindyli (changes requested):
Please test across browsers. Thanks.

src/js/workshop-comments.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {String}[Optional] displayType - CSS display value to apply. Only applicable when toShow is set to true.
+ */
+function applyVisibility(domSelector, toShow, displayType) {
+	displayType = toShow ? displayType : "none";
+	domSelector.setAttribute("style", `display:  ${displayType};`);
+}
+
+/**
+ * Display the new comment on the current page.
+ * @param {String} name - DOM element of name field in comment form.
```
cindyli, line 91: These arguments are no longer DOM elements. Please fix code comments at line 91, 92.

cindyli:
The "post a comment" button still looks broken with Safari on smaller screens such as iPhone SE or iPhone 8.

Screen Shot 2020-09-15 at 9 09 15 AM

cindyli (approved):

Issue #381, opened 2020-08-30
the Email link is gone to next line when we increase space and changes text-style in mobile 


Raunaksingh100:
## Describe the bug
 the email link is gone to next line when we increase the space to """2""" and changed the text-style to """" verdana""" 
see in the screenshot for the better understanding
## To Reproduce

1. Go to 'https://wecount.inclusivedesign.ca/ '   or  'https://dev--wecount.netlify.app/'
2. chaned the text-style to "verdana" and add the space in it "2"
3. Scroll down to 'Footer'

### Screenshots ---
![issue](https://user-images.githubusercontent.com/65535360/91655870-b8571280-ead1-11ea-99a3-9e68fc610cd7.PNG)

## Smartphone (please complete the following information):

- Device: [e.g. galaxy fold]
- Browser [e.g. stock browser]

Raunaksingh100:
### i am working on this issue and you can see this bug in both sites given below--
1. https://dev--wecount.netlify.app/
2. https://wecount.inclusivedesign.ca/

cindyli:
@Raunaksingh100 as we've discussed before with our designer Uttara at #256, it's acceptable that the email gets wrapped into multiple lines.

Closed.

Raunaksingh100:
> @Raunaksingh100 as we've discussed before with our designer Uttara at #256, it's acceptable that the email gets wrapped into multiple lines.
> 
> Closed.

But this bug comes in when we added spaces 
But my code fixed this

PR #382, opened 2020-08-31
aligned the gmail link fixed(#381)

Raunaksingh100:

## Description
the gmail link is gone to next line when add space to "2" and add the text-style"verdana",,,
so i fixed and do the gmail link in one line see in the screenahot for the better understanding...

## Steps to test

1. Go to deploy preview
2. changed text-style to "verdana" and add space to "2"
3. scroll down to footer gmail link

### screenshot after my code changes--
issue fixed we count

### Screenshots before---
issue

## Additional information
gmail link should be aligned in one line as the figma designs

## Related issues
#381

### DIFF
--- a/src/scss/layout/_footer.scss
+++ b/src/scss/layout/_footer.scss
@@ -68,6 +68,7 @@ footer {
 			a {
 				color: $white;
 				font-weight: $font-weight-normal;
+				white-space: nowrap;
 
 				svg {
 					padding: rem(32) rem(16) rem(16) 0;
--- a/src/scss/base/_base.scss
+++ b/src/scss/base/_base.scss
@@ -125,7 +125,3 @@ button::before:not(.fl-prefsEditor-buttons) button {
 		}
 	}
 }
-
-.primary-nav a {
-	border-radius: 20px;
-}
--- a/src/scss/base/_base.scss
+++ b/src/scss/base/_base.scss
@@ -125,3 +125,9 @@ button::before:not(.fl-prefsEditor-buttons) button {
 		}
 	}
 }
+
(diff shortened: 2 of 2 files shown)


src/scss/layout/_footer.scss
```diff
@@ -68,6 +68,7 @@ footer {
 			a {
 				color: $white;
 				font-weight: $font-weight-normal;
+				white-space: nowrap;
```
cindyli, line 71: Although #381 is not an issue and a fix is not needed, just for your future reference, this solution of using `white-space: nowrap` is not ideal.

If you turn on these UIO settings with the "Galaxy Fold" screen size:

1. set "text style" to "Open Dyslexic";
2. Turn on "Enhance inputs";
3. Turn on "Syllables".

This `nowrap` change will force the email stretch out of the viewport and cause the horizontal overflow.

Your change seems working for a specific scenario but it actually introduces a more serious problem when experimenting with more scenarios. Even with the specific scenario you described, this change is too data driven that only works with the current length of an email address. What if the email is changed to a longer email, or what if a smaller screen appears in the market?

Whenever possible, it's preferred that designs and technical solutions are able to accommodate future changes. In this case, a better solution is to allow the email/text to display in multiple lines rather than forcing it to fit into one line.

The problem with this solution is a generic problem with all solutions you issued with WeCount project: you are too focus on fixing an issue with the current data on the page rather than thinking broadly of how your solutions could work in general regardless of data changes and setting changes etc. 

Please stop filing issues and pull requests before you understand what proper solutions are meant to be. Thanks.

cindyli:
Closed as #381 is not an issue.

PR #396, opened 2020-09-16
Fade out comment form while disabled (resolves #395)

### LINKED ISSUE #395: Fade out comment form while disabled in UIO themes

While UIO themes are turned on the background colour of the input fields does not change.

The opacity of the comment form will change while disabled instead of the input fields background colour.

One solution is to apply a lower opacity value to comment form, such as setting it to 0.7 when a UIO theme is selected.

## Additional context

N/A

TeddTech:

## Description
This PR applies a lower opacity value to comment form while disabled.

## Steps to test

1.  Got to Initiatives page of deploy preview
2. Throttle network so it is slow then post a comment

**Expected behavior:** 

Comment form should be faded out while sending post request.

## Additional information
N/A

## Related issues
closes #395


TeddTech:
@cindyli ready for review

cindyli:
@TeddTech, I'm not sure if it's a good indicator to lower the opacity of the entire form including field labels when what was disabled was two input fields. This behavior isn't consistent with what happens when a UIO theme is not selected: grey out just those input fields.

Can you please discuss this solution with @uttaraghodke or @lliskovoi? Thanks.

src/js/workshop-comments.js
```diff
@@ -124,15 +124,19 @@ function disableFields(disabled) {
 	// When a UIO theme is selected, the background color will remain the same as what is set by the theme.
 	const backgroundColor = disabled ? "#f3f3f3" : "#fff";
 
-	// Disable/enable and set background color of post a comment button
+	// Disable/enable and set background color of post a comment button.
```
cindyli, line 127: Add quotes around "post a comment" so this sentence doesn't confuse us after a few months.

src/js/workshop-comments.js
```diff
@@ -124,15 +124,19 @@ function disableFields(disabled) {
...
 	const postCommentButton = document.getElementById("post-comment");
 	postCommentButton.setAttribute("style", "background-color: " + backgroundColor);
 	postCommentButton.disabled = disabled;
+	// When UIO themes are turned on the opacity of the post a commnet button will change while disabled instead of background color.
```
cindyli, line 131: 1. "When a UIO theme is turned on, ...": add comma;
2. Same as the previous comment, add quotes around "post a comment".

Please make sure all your comments still make sense when you or other developers look at them after some time.

src/js/workshop-comments.js
```diff
@@ -124,15 +124,19 @@ function disableFields(disabled) {
 	// When a UIO theme is selected, the background color will remain the same as what is set by the theme.
 	const backgroundColor = disabled ? "#f3f3f3" : "#fff";
 
-	// Disable/enable and set background color of post a comment button
+	// Disable/enable and set background color of post a comment button.
 	const postCommentButton = document.getElementById("post-comment");
 	postCommentButton.setAttribute("style", "background-color: " + backgroundColor);
```
cindyli, line 129: The logic of line 129- 132 and line 137-140 are pretty much the same, please create a function to share.

src/js/workshop-comments.js
```diff
@@ -124,15 +124,19 @@ function disableFields(disabled) {
...
 	postCommentButton.disabled = disabled;
+	// When UIO themes are turned on the opacity of the post a commnet button will change while disabled instead of background color.
+	disabled ? postCommentButton.classList.add("disable-form") : postCommentButton.classList.remove("disable-form");
```
cindyli, line 132: This line can be simplified to:

```
postCommentButton.classList[disabled ? "add" : "remove"]("disable-form");
```

src/scss/abstracts/_mixins.scss
```diff
@@ -175,6 +175,7 @@
 	}
 
 	// Workshop Page
+	.disable-form,
```
cindyli, line 178: This class no longer applies to a form. Rename "disable-form" to something like "disabled-element".

cindyli (changes requested):

TeddTech:
@cindyli ready for review

src/js/workshop-comments.js
cindyli: Thanks for creating `reduceOpacity()` function. But what I meant was to share the entire logic of disabling rather than sharing one single action. Looking further into the code, it's even unnecessary to use separate blocks to disable a button and input fields. Line 136-150 can be simplified to:

```
	const commentFormElms = document.querySelectorAll("#name, #comment, #post-comment");
	commentFormElms.forEach( element => {
		element.setAttribute("style", "background-color: " + backgroundColor);
		const elmType = element.nodeName.toLowerCase();
		element[elmType === "button" ? "disabled" : "readOnly"] = disabled;
		// When UIO themes are turned on, reduce the opacity of form elements as a disabled indicator instead of changing the background color.
		// The css class "disable-element" is only defined with UIO themes.
		element.classList[disabled ? "add" : "remove"]("disable-element");
	});
```

cindyli (changes requested):
See https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/396#discussion_r490258785

PR #415, opened 2020-09-30
feat: host Antranig's initial map demo on the WeCount website

cindyli:

## Description

Host @amb26's initial map demo on the WeCount website.

## Additional information

A couple of remaining issues to be addressed later:

1. @amb26 and I talked about having a html template in [Antranig's map repo](https://github.com/amb26/fluid-covid-map-viz) holding the markup for rendering the map. As this file isn't yet created in there, this file is sourced from this repo at [`src/_includes/components/map.html`](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/add-initial-demo/src/_includes/components/map.html)

2. The future map component needs to be able to fetch a remote data file from [the COVID data repo]. See [the code comment](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/add-initial-demo/src/map.njk#L31-L34) that describes the issue with using a remote file.


cindyli:
@amb26, @ptcher, @TeddTech: [the deploy preview of this PR](https://deploy-preview-415--wecount.netlify.app/map/) demonstrates Antranig's map demo.

amb26:
This pull request doesn't use any of the build files which I produced - here is an "all in one" JS build at https://github.com/amb26/fluid-covid-map-viz/blob/master/build/js/fluid-covid-map-viz.js etc. CSS as referenced in the build test file at https://github.com/amb26/fluid-covid-map-viz/blob/master/build/index.html - 
The html template to be included can be referenced using an njk include at https://github.com/amb26/fluid-covid-map-viz/blob/master/build/html/template.html

amb26:
Perhaps one option for the remote data fetch might be to produce an extremely tiny lambda function that proxies the file from the data repo?

cindyli:
@amb26, I ran into a problem when switching to use the built js file (https://github.com/amb26/fluid-covid-map-viz/blob/master/build/js/fluid-covid-map-viz.js):

1. WeCount pages already includes [infusion-uio.min.js](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/add-initial-demo/src/_includes/layouts/base.njk#L20) in the base template for UIO. It's now using the infusion build from your latest FLUID-6145 branch;
2. `build/js/fluid-covid-map-viz.js` also contains the same infusion version;
3. When infusion is included twice, [uio.js in the WeCount](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/add-initial-demo/src/js/uio.js#L4) throws this error: `Cannot make component creator for type fluid.uiOptions which does not have any gradeNames defined`;
4. Only including the infusion once will fix the problem.

Do you have an idea of why? It feels to me that including the infusion library more than once on a page should be fine. Do I write something wrong in `uio.js`?

amb26:
Hi there @cindyli - thanks for working through this issue! Including Infusion more than once "should" be fine, but it isn't - in practice, the 2nd Infusion to load will blow away the defaults store when it reaches this line - https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/Fluid.js#L1834 - and given the build for my viz doesn't include UIO, the grade is never restored. There are some long-standing notes on how we need to modularise Infusion better at https://issues.fluidproject.org/browse/FLUID-5579 - in practice there should be a separate micromodule named something like "fluid-store" holding the defaults store.
In the meantime, the best thing to do might be for me to make a fresh "Infusionless build" of the viz - this sort of thing is quite common, since I often find myself needing to make a "jQuery-less" build of things

cindyli:
> Perhaps one option for the remote data fetch might be to produce an extremely tiny lambda function that proxies the file from the data repo?

@amb26, I wrote [a netlify function](https://github.com/cindyli/wecount.inclusivedesign.ca/blob/add-initial-demo/functions/getMapData.js) that serves an API `/api/getMapData` to return the covid csv data in a string. However, using it as `dataUrl` on `fluid.mapviz.covid.map`component throws this error: `Failure loading resources for component at path map-2a4bkf2t-38`.

amb26:
I think the issue is that you have incorrectly encoded the payload returned by getMapData. It appears to have undergone a further layer of JSON-encoding - that is, encoding the CSV file as a JSON-encoded string. Evidence for this is that the rendered response from the server begins
````
"active,location_id,location_name
````
and the error thrown by the local CSV encoder is 
````
15:54:42.623:   Failure loading resources for component at path map-22tusbiu-38:  
code: "InvalidQuotes"
index: 1
message: "Trailing quote on quoted field is malformed"
row: 0
type: "Quotes"
````

cindyli:
> I think the issue is that you have incorrectly encoded the payload returned by getMapData. 

Good catch, @amb26. Removing the JSON encoding fixes the problem. Thanks.

One last task is I'm expecting an Infusionless JS build from your map repo. Once I switch to it, this PR should be complete.

cindyli:
Adjusted to use the Infusionless JS build from fluid-covid-map-viz@0.02. 

Once the data repo structure is confirmed(currently a work-in-progress via https://github.com/inclusive-design/covid-assessment-centres/pull/4), the server script serving /getMapData endpoint needs to adjusted to use the new structure of the data repo. Other than that, this PR is ready for now.

PR #423, opened 2020-10-13
Feat: create the Advisory Panel page (resolves #417)

### LINKED ISSUE #417: Create Advisory Panel page

Create the Advisory Panel page. The design of it can be found [here](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website)

The content creator of this page creates a new "Advisory Panel" page in WordPress. The wecount website fetches the page and restyle it to the design.

The content creator of this page creates each advisors as a post in WordPress. The wecount website fetches individual post records and assemble them into the Advisory Panel page.

cindyli:

## Description

Implement "Advisory Panel" page.

## Steps to test

1. Go to the advisory panel page at {domain}/advisory-panel;
2. The function and style should match [the figma design](https://www.figma.com/file/0lcLol3X5MmOXackT2YbHJ/WeCount-website?node-id=2257%3A0);
3. Test with UIO and everything should work well.

**Expected behavior:** 

See above.


TeddTech (commented):
In Jutta Treviranus's bio the following should be a hyperlink for "Inclusive Design Research Centre (IDRC)"

Screen Shot 2020-10-13 at 12 49 25 PM

TeddTech (commented):
@uttaraghodke should the background of the bio cards be white when there is no UIO theme turned on (default setting)? As of now they are clear.

Screen Shot 2020-10-13 at 2 35 06 PM

TeddTech (commented):
Screen Shot 2020-10-13 at 2 31 51 PM

Expansion icon for card bios overlaps cards when UIO text style is set to Verdana and enhance inputs are turned on in mobile views.

cindyli:
> In Jutta Treviranus's bio the following should be a hyperlink for "Inclusive Design Research Centre (IDRC)"
> 
This is from Jutta's bio document. 

@uttaraghodke, I will leave it to you to decide if Jutta's bio should be updated. Thanks.

cindyli:
@uttaraghodke and I discussed comments above that require her input:

1. Bio cards should use white background on the mobile view;
2. Jutta's bio is a placeholder at the moment. It will be replaced once Jutta provides her bio.

@TeddTech, all comments have been addressed. This pull request is ready for another review.

TeddTech (commented):
In Galaxy Fold view the background color of bio cards are not aligned.

Screen Shot 2020-10-15 at 1 05 12 PM

cindyli:
@TeddTech, the latest comment has been addressed by removing the special css handling for the galaxy fold screen. I figure we should rely on soft hyphens for displaying a long word in multiple lines, rather than adding a special handling which seems leading to more special handling for fix issues resulted from it.

Ready for another look. Thanks.

src/scss/components/_advisory-panel.scss
```diff
@@ -1,6 +1,7 @@
 // CSS classes for the "Advisory Panel" page
 .advisory-panel {
 	.api-content > .wp-block-group {
```
TeddTech, line 3: Only add the white background to the bio cards instead of having all the content in `.api-content` have a white background.

cindyli: > Only add the white background to the bio cards instead of having all the content in `.api-content` have a white background.

Good catch. Fixed and ready for another look.

TeddTech: Looks good. After making that change it seems like re-adding the special css handling for the galaxy fold screen will work as intended now or we can leave it out and continue with relying on the soft hyphen solution.

cindyli: I lean towards relying on the soft hyphen solution which applies to all cases. It'll reduce the maintenance work by getting rid of special handlings.

TeddTech (approved):

PR #468, opened 2021-03-31
feat: configure Netlify CMS for workshops

greatislander:

## Description

This PR migrates workshops from Airtable into an Eleventy file-based collection, and configures a Netlify CMS instance to manage the collection.

## Steps to test

1. Review the Initiatives page on the deploy preview.
2. Check out the branch, run `npm start`, and navigate to http://localhost:3000/admin.

**Expected behavior:**

1. The Initiatives page and individual initiatives are properly displayed.
2. The workshops are editable via the CMS interface with live previews.

## Additional information

See updated [docs](https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/0e0840886c2ca7385fdb2aadc1570ac6c08ae4cc/README.md#content-management-system-cms).

Note that the CMS is set to manage the `dev` branch currently. This will need to be changed once this PR has been tested and is merged to `main`.

## Related issues

Not applicable.


src/admin.njk
```diff
@@ -0,0 +1,25 @@
...
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Content Manager</title>
+  <script src="https://identity.netlify.com/v1/netlify-identity-widget.js"></script>
+</head>
+<body>
+  <!-- Include the script that builds the page and powers Netlify CMS -->
+  <script src="https://unpkg.com/netlify-cms@2.10.103/dist/netlify-cms.js"></script>
+	{% if env.context === "dev" %}
```
cindyli, line 15: Please sync up the use of tabs and spaces for indentation. Line 15-21 use tabs for indentation while other lines use 2 spaces. Please use tabs across.

This comment applies to other files too. Thanks.

src/comments.njk
```diff
@@ -0,0 +1,8 @@
+---
+permalink: /comments.json
```
cindyli, line 2: The generated `comments.json` doesn't seem to be used anywhere.

README.md
```diff
@@ -39,6 +43,107 @@ From the root of the cloned project, enter the following in your command line to
...
+        name: "id"
+        widget: "uuid"
+        hint: "The ID is used to associate comments with this workshop and cannot be edited."
+      - label: "Permanent Link"
+        name: "permalink"
+        widget: "string"
+        required: false
+        hint: |-
+          If you do not specify a permanent link, one will be automatically generated from the event title.
+          Permalinks must have the format /initiatives/event-title/ (the trailing slash is required).
```
cindyli, line 80: Shall the format of permalinks be /initiatives/event-id/? event-title -> event-id

The same comment applies to `src/admin/config.yml` where the hint for the "permalink" field is defined.

src/admin/config.yml
```diff
@@ -0,0 +1,59 @@
+backend:
+  name: git-gateway
+  branch: dev
```
cindyli, line 3: As we talked yesterday in the channel, the plan is to use different config files for `dev` and `main` branches. Do you have thoughts about how to handle this difference during the PR merge process.

cindyli:
I noticed when an image is uploaded from the local disk, the preview panel won't be able to show the image. If the image is specified via an URL, the preview responds to this change. This bug appears to be [a known issue](https://github.com/netlify/netlify-cms/issues/3158) with Netlify CMS.

src/collections/workshops/workshops.json
```diff
@@ -0,0 +1,4 @@
+{
+	"permalink": "/initiatives/{{ title | slug }}/",
```
cindyli, line 2: You taught me the other day that even though the permalink is defined as using `title` in here, it will be overridden later with the even id. I forgot where the overriding is.

I wonder if it would be more straightforward to directly use `id` rather than `title` here without relying on the overriding?

greatislander: The permalink can be overridden in an individual workshop file (see the value defined for `permalink` [here](https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/d462045ce51890ec82301f7a39af8503ed4b8230/src/collections/workshops/recQcaxqWsxtnljqe.md)) or via the CMS `permalink` field as per https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/468#discussion_r605882949.

I think it's a good idea to default to titles going forward (it makes for friendlier, more readable URLs), but since these workshops have been online for a while and probably linked to in various newsletters etc, I am manually overriding the permalink with the event ID so that the links do not break.

README.md
greatislander: See comment below: https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/468#discussion_r605910729

greatislander:
> I noticed when an image is uploaded from the local disk, the preview panel won't be able to show the image. If the image is specified via an URL, the preview responds to this change. This bug appears to be [a known issue](https://github.com/netlify/netlify-cms/issues/3158) with Netlify CMS.

I think this is actually an unrelated issue. But that's a good reminder that we need to test image uploading once this is merged into dev and make any necessary adjustments before it goes into main, and also that I need to move the images for existing workshops from Airtable to the uploads path in the images folder here.

src/admin/config.yml
greatislander: My understanding of git merging is that if we:

1. Merge this to dev.
2. Merge dev to main.
3. Update main's config.yml to point to main instead of dev.

Future merges from dev won't update that line since it will have been modified more recently in main than in dev. We'll need to verify this behaviour though.

src/collections/workshops/workshops.json
cindyli: So the new workshop records will use `title` in permalinks. I agree using titles is more readable and friendlier.

greatislander: That's right— and if we want we can also change the existing URLs and add redirects. But I thought it would be good to keep this as straightforward as possible without changing URLs as part of the new CMS setup.

greatislander:
@cindyli I made the change from Workshops to Initiatives. If the build passes this is now ready to merge to dev.

cindyli:
Thanks for the renaming, @greatislander.

The other change we talked at today's meeting was to add the "Event Date" to the "sort by" list for the initiatives collection. It seems this can be achieved by using [the `sortable_field` config](https://www.netlifycms.org/docs/configuration-options/#sortable_fields) in the config.yml.

greatislander:
@cindyli I've added the configuration.

cindyli:
The sorting works great, @greatislander.

I checked the initiatives data in the Airtable and this PR. It seems there's at least changes to the "The Metric Society and the Unmeasurable" workshop since your last data update. Could you re-pull the Airtable data before I merge this PR? Thanks.

greatislander:
@cindyli I've updated the content from Airtable.

src/collections/initiatives/recM2F4PCtF3ufqZ3.md
```diff
@@ -1,17 +1,21 @@
...
-  https://dl.airtable.com/.attachments/1f23e46ecca1375046369ab8229ddb0a/e7d8f19f/group82.jpg
-previewImageAltText: ''
-coverImageUrl: >-
-  https://dl.airtable.com/.attachments/f5d282fcac73b5359a4562c21c560c9b/88950a55/AccessibleSurveyPlatforms_1024x536.png
-coverImageAltText: >-
-  Online Conversation on Accessible Survey Platforms. Tuesday, August 4th. 2:00
-  PM - 3:30 PM. Register for free
+permalink: /initiatives/recM2F4PCtF3ufqZ3/
+eventDate: 2020-08-04T03:00:00.000Z
+coverImageUrl: https://dl.airtable.com/.attachments/f5d282fcac73b5359a4562c21c560c9b/88950a55/AccessibleSurveyPlatforms_1024x536.png
```
cindyli, line 6: Shall we download images from Airtable and place in a local directory? When we clean up Airtable, these images won't be accidentally removed.

greatislander: Oh yeah, good call. I'll do that.

src/collections/initiatives/recQcaxqWsxtnljqe.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Inclusive AI for HR
+id: recQcaxqWsxtnljqe
+permalink: /initiatives/recQcaxqWsxtnljqe/
+eventDate: 2020-12-01T04:00:00.000Z
+coverImageUrl: /images/uploads/FWDlogoEforprocessing4x.png
```
cindyli, line 6: There are a few typos in image names in these initiatives markdown files, which causes these images are not shown in the content editing fields. For example, the filename for this image in the local directory is `FWDlogoBforprocessing4x1.png`. Same type of typo for the preview image in the "Coding to Learn and Create" workshop.

Could you go through the initiatives collection and make sure all images work as expected? Thanks.

cindyli (approved):

PR #487, opened 2021-04-08
feat: migrate WordPress pages to Eleventy + Netlify CMS

greatislander:

## Description

This PR addresses the first phase of migration from WordPress: pages. The only exceptions are the News and Views pages, as their integration with the WordPress News and Views sources is very close and it makes more sense to handle that migration independently.

## Steps to test

1. `npm run dev`
2. Review all pages.
3. Review editing workflow of all pages.

**Expected behavior:** All pages are displayed appropriately and are editable via the CMS.

## Additional information

- Re-importing WordPress pages will overwrite some of the new custom content implementations, notably the "Expanders" for Advisory Panel members, YouTube embeds, and image and text widgets on the badges page. As such, it's not advised that the provided migration script be run again.
- This PR implements some new shortcodes and editor widgets to support them. See `src/shortcodes` and `src/admin/cms.js` for examples (docs: [11ty](https://www.11ty.dev/docs/shortcodes/) / [Netlify CMS](https://www.netlifycms.org/docs/custom-widgets/#registereditorcomponent).

## Related issues

Not applicable.


src/utils/data-fetcher-wp.js
```diff
@@ -96,7 +96,7 @@ module.exports = {
 		// According to the Wordpress API for pagination and embedding: https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/,
 		// 1. fetch 100 records per page (the maxium number per page supported by Wordpress) to fasten the query
 		// 2. Order by the menu order in ascending order
-		const pageAPI = env.wpApi + "/pages?per_page=100&order=asc&orderby=menu_order";
+		const pageAPI = "https://wecount-cms.inclusivedesign.ca/wp-json/wp/v2/pages?per_page=100&order=asc&orderby=menu_order";
```
cindyli, line 99: Is this the residue from the migration? Shall it be changed back to replace the hardcoded production API to `env.wpApi`? Thanks.

src/js/expander.js
```diff
@@ -0,0 +1,12 @@
+document.addEventListener("click", event => {
+	if (!event.target.closest(".expander button")) return false;
```
cindyli, line 2: It would be more efficient to register the click event listener directly to ".expander button". You may need to register the click event within the `window.load` event listener to make sure the page has been loaded and buttons are available for hooking up the click event.

package.json
```diff
@@ -28,13 +28,16 @@
     "lint-fix:js": "npm run lint:js -- --fix",
     "lint-fix:md": "npm run lint:md -- --fix",
     "lint-fix": "run-s lint-fix:*",
+    "migrate:wp": "run-s build wp:*",
```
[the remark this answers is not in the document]
greatislander: I'd like to keep this in dev until all the WordPress migration is done as it shows a method of importing content from WordPress.

src/utils/data-fetcher-wp.js
greatislander: I don't think there's any need to change it back; rather, once you're done reviewing this and I've addressed any changes I can remove the wpPages fetcher altogether.

src/js/expander.js
greatislander: @cindyli Just to clarify, your suggestion is to attach a click event listener to all expander buttons once window has loaded? Thanks!

cindyli: Yes, @greatislander.

greatislander: (I was following this approach, slightly modified to make sure we catch clicks on the `<svg>` within the button: https://gomakethings.com/listening-for-click-events-with-vanilla-javascript/#the-vanilla-js-way-to-listen-for-click-events).

cindyli: This article said: `While you can listen for clicks on specific elements, my recommended approach is to listen for all clicks on the document, and then check if the clicked element has the selector you care about.`

I'm not sure what's the of doing it while specific elements are known and listen to the click event directly. It feels to me the suggested approach is less efficient.

greatislander: @cindyli The author has a follow-up explanation which is interesting: https://gomakethings.com/why-event-delegation-is-a-better-way-to-listen-for-events-in-vanilla-js/

src/collections/pages/badges.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Badges
+permalink: /badges/
+slug: badges
+excerpt: -| We Count badges enable earners to showcase their proficiency in the
+  growing fields of AI, data systems and inclusive data practices as well as
+  other skills. Badges, once…
+menu_order: 0
+---
+![Horizontal rectangle banner of four badges of WeCount. Learner, brainstormer, collaborator and communicator. All the badges are hexagonal in shape.](/uploads/banner-06-1024x270.jpg)
```
cindyli, line 10: A typo in the image name: jpg -> jpeg.

cindyli:
What are the steps to point a link to a uploaded non-image file such as pdf?

src/js/expander.js
cindyli: It's an interesting argument. I agree this approach helps to write less code but I think it's bad for the performance and the code logic.

cindyli:
A couple of issues with using "image and text" editor component on CMS:

1. When an image is uploaded from the local drive, the preview doesn't show this image;
2. The image position "left" doesn't work: the image is displayed on the top of the content;
3. When the content is short in less than a line, the preview of the content is too close to the image. See the screenshot:
close

cindyli:
The "expander" editor component has the same issue when the image is uploaded from the local drive: the preview doesn't show this image.

greatislander:
> The "expander" editor component has the same issue when the image is uploaded from the local drive: the preview doesn't show this image.

This is now fixed in 78a7de5 for both block types.

greatislander:
> A couple of issues with using "image and text" editor component on CMS:
> 
> 1. When an image is uploaded from the local drive, the preview doesn't show this image;

Fixed!

> 2. The image position "left" doesn't work: the image is displayed on the top of the content;

I looked at the styles and it appears that this is an existing bug in the CSS unrelated to these changes. I think this hasn't been noticed because there aren't any image and text blocks on the site that have the image positioned on the left. Can you open a ticket if this is something that should be supported? Otherwise I can remove the left position option from the editor.

> 3. When the content is short in less than a line, the preview of the content is too close to the image. See the screenshot:
> 
> close

I believe this is also an issue with the existing site—can you open a separate ticket to track it?

greatislander:
> What are the steps to point a link to an uploaded non-image file such as pdf?

Still trying to figure out the best approach for this…

src/js/expander.js
```diff
@@ -1,12 +1,14 @@
...
-	button.setAttribute("aria-expanded", !expanded);
-	if (expanded) {
-		content.setAttribute("hidden", "true");
-	} else {
-		content.removeAttribute("hidden");
-	}
+window.addEventListener("load", () => {
+	document.querySelectorAll(".expander button").forEach(button => {
+		button.addEventListener("click", () => {
+			const expanded = button.getAttribute("aria-expanded") === "true" || false;
```
cindyli, line 4: The part of `|| false` is unnecessary. The condition in front will take care of this part.

cindyli:
> > 1. The image position "left" doesn't work: the image is displayed on the top of the content;
> 
> I looked at the styles and it appears that this is an existing bug in the CSS unrelated to these changes. I think this hasn't been noticed because there aren't any image and text blocks on the site that have the image positioned on the left. Can you open a ticket if this is something that should be supported? Otherwise I can remove the left position option from the editor.
> 
> > 1. When the content is short in less than a line, the preview of the content is too close to the image. See the screenshot:
> 
> I believe this is also an issue with the existing site—can you open a separate ticket to track it?

Make sense. I created https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/488 and https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/489 for them.

src/shortcodes/image-and-text.js
```diff
@@ -1,7 +1,8 @@
 const markdownFilter = require("../filters/markdown.js");
 
 module.exports = (content, image, alt, imagePosition, verticalAlignment) => {
-	return `<div class="wp-block-media-text has-media-on-the-${imagePosition} is-vertically-aligned-${verticalAlignment} is-stacked-on-mobile">
+	const imagePositionClass = imagePosition === "right" ? "has-media-on-the-right " : "";
```
cindyli, line 4: If "left" is supported, it's probably better to remove it from the editor component so that we can keep your old code of `has-media-on-the-${imagePosition}`, which gives the flexibility of adding the support of "left" in the futre.

BTW, `is-stacked-on-mobile` class is not styled. Can we take the chance to remove it?

cindyli (approved):

PR #940, opened 2022-01-25
feat: move from sanitize to modern-css-reset (resolves #939)

### LINKED ISSUE #939: Move from use of sanitize for CSS normalisation to "modern CSS reset"
Following discussion in the Fluid Design Matrix chat at https://matrix.to/#/!CiNvDHGGdtIVCpYoGy:matrix.org/$3grOKZ9M5aKm49orKwJwD5STiOaIcAN0Jb2E5ssoPrs?via=matrix.org it was determined we should move from the historical choice of sanitize.css to a more minimal library - initially prompted by the "opinionated" CSS rule that overrides the browser's default choice of cursor which may be confusing for some users.

The original evaluation for a normalisation library was done at https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/108 but the rationale for choosing sanitize.css was not recorded.

@greatislander reported that many of our projects are moving away from more heavyweight CSS normalisation, and in particular Infusion itself moved away from normalise.css in https://issues.fluidproject.org/browse/FLUID-6607. He suggested the use of a more minimal, modern library such as https://piccalil.li/blog/a-modern-css-reset/ (in npm as https://www.npmjs.com/package/modern-css-reset ) as a good alternative.

amb26:

## Description

Migrate from CSS normalisation library sanitize.css to "modern CSS reset"

**Expected behavior:**

Hovering over any text element should show the standard "caret" cursor indicating selectable text

## Related issues

Resolves #939

### DIFF
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -83,10 +83,8 @@
 			margin-top: rem(24);
 
 			button {
-				align-items: center;
 				background-color: $white;
 				box-shadow: 0 0 rem(7) $border-shadow-colour;
-				display: flex;
 			}
 		}
 
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -83,6 +83,8 @@
 			margin-top: rem(24);
 
 			button {
+				@extend .has-centred-children;
+
 				background-color: $white;
 				box-shadow: 0 0 rem(7) $border-shadow-colour;
 			}
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -99,6 +99,8 @@
 			}
 
 			button {
+				@extend .has-centred-children;
+
 				align-self: center;
 				height: rem(32);
 				margin-right: 0;
@@ -130,6 +132,8 @@
 		}
 
 		.filter-expand-button {
+			@extend .has-centred-children;
+
 			--expand-button-color: none;
 			--expand-button-stroke-color: #{$grey-dark};
 
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -83,8 +83,10 @@
 			margin-top: rem(24);
 
 			button {
+				align-items: center;
 				background-color: $white;
 				box-shadow: 0 0 rem(7) $border-shadow-colour;
+				display: flex;
 			}
 		}
 
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -83,7 +83,7 @@
 			margin-top: rem(24);
 
 			button {
-				@extend .has-centred-children;
+				@extend %has-vertically-centered-children;
 
 				background-color: $white;
 				box-shadow: 0 0 rem(7) $border-shadow-colour;
@@ -101,7 +101,7 @@
 			}
 
 			button {
-				@extend .has-centred-children;
+				@extend %has-vertically-centered-children;
 
 				align-self: center;
 				height: rem(32);
@@ -134,7 +134,7 @@
 		}
 
 		.filter-expand-button {
-			@extend .has-centred-children;
+			@extend %has-vertically-centered-children;
 
 			--expand-button-color: none;
 			--expand-button-stroke-color: #{$grey-dark};
--- a/src/scss/layout/_layout.scss
+++ b/src/scss/layout/_layout.scss
@@ -38,6 +38,12 @@ article {
 	}
 }
 
+// style for centering children of some buttons (#939)
+.has-centred-children {
+	align-items: center;
+	display: flex;
+}
+
 // WP customized styles
 .has-small-font-size {
 	font-size: 84%;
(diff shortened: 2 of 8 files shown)


cindyli:
It seems the new reset library handles heading margins, button widths differently. A few things I noticed on the resources page:

1. The gaps between filter section headers are reduced.
2. The alignment of the SVG and the button label on buttons is broken.

The screenshot of before:
before

The screenshot of after:
after

cindyli:
Also on the resources page, 

1. The margin around the page title "Resources" is reduced.
Before:
before-title

After:
after-title

2. On each resource tile, he margin between the upper border of the tile and the resource title is reduced. 
Before:
before

After:
after

These are probably also because the new reset library doesn't apply margin to heading elements.

cindyli:
The margin issue mentioned above can be found on every page. Headings, paragraphs and texts are more squeezed together.

greatislander (commented):
I took a quick look and this looks good to me, but I will defer to @cindyli for approval on this :)

src/scss/components/_resources.scss
```diff
@@ -83,8 +83,10 @@
 			margin-top: rem(24);
 
 			button {
+				align-items: center;
```
cindyli, line 86: It might be better to create a generic `button` style in `_base.scss` with the adjustment migrated from sanitize so it applies to all buttons across the site. There are more buttons on "Initiatives" page and "Events" page.

amb26: I wasn't able to find the reason in detail why the children of the filter buttons ended up centre aligned under sanitize and so I thought it would be best to be conservative with this change and only apply it here. The buttons on the initiatives and events page aren't affected by this layout rule since they are simple buttons without child elements.

amb26: Sorry, I see your point now - the other buttons have children which are the expand and contract arrows which are also slightly misaligned because of this change in layout rule. I've applied this suggestion, thanks.

amb26:
Cheers, ready for another look - I think with the new button alignment rule the layout looks even better than before.

cindyli:
Thanks for the fix, @amb26. It looks great.

Another issue I just noticed is, when clicking "show preferences" to open UIO, the display of "hide preferences" and "reset" buttons breaks into 2 lines rather than remaining in the same line. This seems like a side effect of using `display: flex` on buttons.

Screen Shot 2022-02-01 at 9 02 22 AM

cindyli:
> This seems like a side effect of using `display: flex` on buttons.
> 
After some thought about this issue, when it comes to use `display: flex` to align children elements in a button, your idea of only applying it to individual buttons make more sense.

cindyli:
Buttons on the initiatives page need to be centered too:
Screen Shot 2022-02-08 at 1 42 33 PM

src/scss/layout/_layout.scss
```diff
@@ -38,6 +38,12 @@ article {
 	}
 }
 
+// style for centering children of some buttons (#939)
+.has-centred-children {
```
cindyli, line 42: A couple of thoughts about defining this general class:

1. Rules in here only center the first layer of children elements vertically. It won't affect children in deeper layers and also doesn't center horizontally, so the name is a little misleading.
2. There are side effects to use `display: flex` to centre as we saw what happened to UIO buttons with this rule. Whether to use flex box to center is a case by case decision depending on the context.

As this rule is not general enough, I personally prefer to directly add these rules when they are needed rather than defining it as a general class.

I'd love to hear thoughts from you and @greatislander.

amb26: The rules are added where they are needed - via the "extends" directive that I added in _resources.scss, etc. 
I think making the name even longer than it currently is to account for the fact that only direct children are centred would be a bit cumbersome, but I'm open to suggestions.

greatislander: A couple thoughts on this. The `.has-centered-children` class only centers items along the horizontal axis, as `display: flex` without `flex-direction` is implicitly `row` (as @cindyli noted). So it might be more precise for the extended class to be `.has-vertically-centered-children`. I'm less concerned about the use of "children" (if it were `.has-centred-descendents` I would be more concerned but I think of children as direct descendents only).

An alternative to defining this as a class and extending it would be a Sass placeholder selector. [Here are the docs](https://sass-lang.com/documentation/style-rules/placeholder-selectors). Essentially, it can still be extended but won't show up as an independent class in the compiled CSS:

Sass:
```scss
%has-vertically-centered-children {
    align-items: center;
    display: flex;
}

.filter-expand-button {
    @extend %has-vertically-centered-children;
}

.filter-header {
    button {
        @extend %has-vertically-centered-children;
    }
}
```

CSS:

```css
.filter-header button, .filter-expand-button {
  align-items: center;
  display: flex;
}
```

This makes the output a little cleaner.

As one last note I would encourage using `centered` as I misspelled this in my entire comment and then looked back to see that @amb26 had used the UK English spelling `centred` — I think since CSS itself uses `center` we should stick to that for class/utility names.

amb26: Thanks for the great review, @greatislander - all suggestions applied

greatislander (approved):
Looks great @amb26!

PR #1032, opened 2022-06-27
fix: restore bullet styling which was lost in transition to modern-css-reset

amb26:

## Description

As part of our transition to modern-css-reset in https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/940 we lost a few styling rules from our reset, including those which applied default HTML list styling. As a result, unordered lists in markdown appear without any styling as shown in the following screenshot

wecount-bullets

## Steps to test

View a page with a bulleted list such as https://wecount.inclusivedesign.ca/views/pluralistic-data-infrastructure/

**Expected behavior:**

List should appear styled as a default HTML unordered list

### DIFF
--- a/src/scss/base/_base.scss
+++ b/src/scss/base/_base.scss
@@ -63,7 +63,7 @@ a:focus {
 }
 
 li {
-	list-style: none;
+	list-style: unset;
 }
 
 p {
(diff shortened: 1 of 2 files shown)


cindyli (changes requested):
This affects the page navigation bar that now looks like:

Screen Shot 2022-06-27 at 1 08 59 PM

amb26:
Good catch, @cindyli - I found a stray rule in the wrong blog in _pagination.scss that was meant to be ensuring this

cindyli (approved):

PR #1100, opened 2023-01-09
fix: solve Initiatives tags sorting and duplicate tag issues

### LINKED ISSUE #816: Initiatives checkbox cases have ASCII sorting
## Describe the bug

The filter checkbox items on the Initiatives page are sorted in strict ASCII order and have inconsistent case usage.

E.g. "Toronto" is listed before "inclusive"

## To Reproduce

1. Go to the Initiatives page
2. Click on the Filter header to show the checkboxes
3. See error

## Expected behavior

Items should have consistent case usage (e.g. all capitalized) and sorting should not be strictly based on ASCII values.

## Desktop:

N/A

## Smartphone:

N/A

## Additional context

N/A

BlueSlug:

## Description

This update fixes two issues with the Initiatives (formerly Views) page. The first is #816, and the proposed solution was to make all tags lowercase only. The other issue is an undocumented bug which fails builds if two articles had the same tag with different case usage (e.g. `"accessibility"` and `"Accessibility"`).

Seeing that the solution in both cases was to consolidate tags to a single spelling, these are both addressed with a single pull request.

## Steps to test

1. Run the site locally
2. Go to the Initiatives page: http://localhost:3000/views/
3. Expand the "Search Filters" panel
4. Take note of the listed tags.
   Expected state: no duplicate spellings, sort order is alphabetical going left-to-right then top-to-bottom, uppercase and lowercase have no bearing on sort order.
5. Select the "Inclusion" checkbox and activate "Apply Filters"
6. After the matching results are shown (expected: 3 results), click on the link for "We Count Evaluates Toronto's Quayside ..."
7. On the next page, scroll down and click on the link for the "Inclusion" tag
8. On the page for the "Inclusion" tag, there should be six results including ones titled "A Future Date Conference" and "The Walrus Talks: Inclusion" which are news articles.

## Additional information

The initially-proposed fix of making all tags lowercase-only made the multiple acronym tags appear somewhat awkward and more difficult to identify visually. It appeared that getting the site to work by accepting the "lowest" spelling of a given tag rather than enforcing all of them to be lowercase-only would make things significantly more palatable, so this is the approach I took.

Considerations and brief attempts were also made towards better factoring for the fix, including more efficient algorithms and data structures to iterate over the various tags collections, however the time I estimated it would take to design and implement these would far exceed the additional build times created by the current factoring.

## Related issues

Resolves #816


.eleventy.js
```diff
@@ -25,19 +25,27 @@ require("./src/js/utils.js");
...
-	const tagsSet = new Set();
+	let tagsMap = new Map();
 	collection.forEach(item => {
 		if (!item.data.tags) return;
 		item.data.tags
 			.filter(tag => !["pages", "initiatives", "news", "views", "comments"].includes(tag))
-			.forEach(tag => tagsSet.add(tag));
+			.forEach(tag => {
+				if (!tagsMap.has(tag)) {
+					tagsMap.set(tag, slugFilter(tag));
```
cindyli, line 42: Using slugified tags as map keys might be a better option since the slug function will lower case all letters. This will make sure all case related handling afterwards can be removed. Line 41-43 looks like:

```
let slugifiedTag = slugFilter(tag);
if (!tagsMap.has(slugifiedTag)) {
	tagsMap.set(slugFilter(tag), tag);
}
```

Moreover, the sort at line 48 could sort by case-insensitive slugs as well.

.eleventy.js
```diff
@@ -87,41 +72,37 @@ module.exports = function(eleventyConfig) {
 	});
 
 	eleventyConfig.addCollection("allTags", collection => {
-		const tags = getUniqueTags(collection.getAll());
+		let postsByTag = [];
+
+		getUniqueTags(collection.getAll()).forEach(tag => {
+			postsByTag.push({ ... tag, posts: getFilteredByTagTransformed(tag.slug, collection, { transform: slugFilter }) });
```
cindyli, line 78: It seems `getFilteredByTagTransformed` is only used once at here. I would suggest to remove the 3rd `options` argument and directly define and use `slugFilter` in `get-filtered-by-tag-transformed.js`. This will make it easier to read the entire handling in one file rather than coming back here to check what's the transform function.

By doing so, the definition of `slugFilter` at line 19 can be removed from this file.

When the time comes that `getFilteredByTagTransformed` may called more with various transform functions, we could do such refactoring.

cindyli (approved):

PR #1278, opened 2023-06-19
fix: correct service worker caching issue

### LINKED ISSUE #1277: Service Worker caching incorrectly
## Describe the bug

Content or design updates on the website are not always reflected after publishing which would prevent anyone with an older version of the site installed locally from seeing them. The solution is to deregister any active Service Workers in the client browser, at which point fresh copies of the site pages will be fetched and properly displayed.

This is a significant usability issue, for example when hundreds of resources which have been added are simply invisible to those experiencing the issue. Such affected users would have no indication that they are missing updated or additional site content and no way to know without following other channels of communication.

## To Reproduce

Reproducing this bug requires making changes to the site which would require a page reload after publishing. This bug has not been observed outside of the production environment, however I suspect it is also a factor on the [dev preview site](https://dev--wecount.netlify.app/) and in the local development environment.

To reproduce the behavior, publish a change to the production site and observe whether the change is visible upon reloading the page.

## Expected behavior

Site updates are correctly retrieved and displayed once published.

## Screenshots

image

## Desktop:

- OS: Multiple including Windows 10 Pro
- Browser: Multiple including Chrome 104

## Additional context

#833 may benefit from solving this issue.

In my investigations into this issue, I noted that the service worker status, according to Chrome, was stuck at "trying to install". This was observed in production and in my local dev environment (see above).

I discovered an error in `manifest.json` relating to the `purpose` field of the 192px icon. It was set to `any maskable` rather than `maskable` (`manifest.json:15`), which could indeed be the source of the issue (I intend to test this with a PR deploy preview)

In addition to this, I also discovered errors with defining the scope of the Service Worker which I believe directly contributes to it being unable to install due to the `service-worker.js` file not being present.

BlueSlug:
* [ ] This pull request has been linted by running `npm run lint` without errors

## Description & Steps to test

Once complete, this update aims to solve the caching issues related to the service worker not installing properly. Please refer to the description of #1277 for more details.

## Related issues

Resolves #1277

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,4 +1,5 @@
 {
+  "id": "/",
   "name": "We Count",
   "short_name": "WeCount",
   "description": "Removing bias and exclusion in the data economy.",
@@ -11,7 +12,7 @@
     "src": "/images/logo-192.png",
     "type": "image/png",
     "sizes": "192x192",
-    "purpose": "any maskable"
+    "purpose": "maskable"
   }, {
     "src": "/images/logo-512.png",
     "type": "image/png",


greatislander:
Replaced by https://github.com/inclusive-design/wecount.inclusivedesign.ca/pull/1553.

PR #1553, opened 2023-09-18
fix: fix the service worker console error

cindyli:

## Description

Fix https://github.com/inclusive-design/wecount.inclusivedesign.ca/issues/1277. 

## Steps to test

1. Open the console log, reload the page.

**Expected behavior:** 
The error about the service-worker.js is not found should disappear.

### DIFF
--- a/src/_includes/globals/service-worker.njk
+++ b/src/_includes/globals/service-worker.njk
@@ -1,0 +1,6 @@
+<!-- Add service worker -->
+<script>
+  if ("serviceWorker" in navigator) {
+    navigator.serviceWorker.register("/service-worker.js");
+  }
+</script>
--- a/src/_includes/globals/service-worker.njk
+++ b/src/_includes/globals/service-worker.njk
@@ -1,6 +1,0 @@
-<!-- Add service worker -->
-<script>
-  if ("serviceWorker" in navigator) {
-    navigator.serviceWorker.register("/service-worker.js");
-  }
-</script>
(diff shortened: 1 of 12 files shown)


src/_includes/globals/service-worker.njk
```diff
@@ -1,6 +0,0 @@
-<!-- Add service worker -->
-<script>
-  if ("serviceWorker" in navigator) {
-    navigator.serviceWorker.register("/service-worker.js");
-  }
-</script>
```
greatislander, line 6: @cindyli Was your intention to remove the service worker? Now it is being built properly (and appears in the build directory as `service-worker.js`) but by removing this file and the include in `layouts/base.njk` it is no longer being loaded. This will pose a problem as the old service worker will still be registered for visitors who visited before it broke, which means they may still see out of date content. If you want to remove the service worker altogether you would need to follow the steps here: https://github.com/lwojcik/eleventy-plugin-pwa-v2#uninstalling-the-plugin

src/_includes/layouts/base.njk
```diff
@@ -5,9 +5,6 @@
 <html lang="en-ca">
   <head>
     {% include 'globals/matomo-analytics.njk' %}
-    {% if env.context !== "dev" %}
-    {% include 'globals/service-worker.njk' %}
```
greatislander, line 9: As above, I'm not sure why this was removed.

greatislander (changes requested):

src/_includes/globals/service-worker.njk
cindyli: I was intended to upgrade the service worker but didn't understand `service-worker.js` was generated by the updated package in the build directory. :) Thanks for pointing out, @greatislander. It has been fixed and ready for another look.

greatislander (approved):

PR #1616, opened 2024-01-11
feat: update project to follow Trivet conventions

chosww:

## Description

This pull request restructures some parts of the site (assets, etc) to follow Trivet conventions, adds Trivet linting rules, and switches to eleventy-plugin-fluid for asset processing.

## Steps to test

1. Test linting.
2. Test building.
3. Test running in dev mode.

**Expected behavior:** All work as expected.

## Additional information

Not applicable.

## Related issues

#1469, #1470, #1471, #1472


greatislander:
Resolved navigation issue. However, lots to do still:

1. Resources index currently only includes English resources; so filtering won't work for French resources.
2. Resource types, tags, etc. are not translated into French but need to be.
3. Search likely won't work for French content. I think this is probably a good time to explore using https://pagefind.app for searching.

chosww:
Hey @greatislander, thanks for resolving the navigation issue. I was looking into localization work for resources related global data, and I have an idea but couldn't make it work using Decap CMS. I started by updating resource datas to have each locale attribute, for example, resourceFocuses.json would have new structure:
```
{
  en-CA: {
    resourceFocuses: [
    ]
  },
  fr-CA: {
    resourceFocuses: [
    ]
  }
}
```

This new structure works when I specify locale in config.yml, like en-CA.resourceFocuses.*.value; however, because I hardcoded the locale, it isn't working for fr-CA. I tried to google how to get the correct locale, but didn't have luck. Do you think the approach I am thinking is correct? Would be great to discuss about this.

chosww:
@greatislander For localization work for Pages collection, keeping all supported language files in pages directory makes sense? With the current structure, we are just differentiate different locale files by adding locale to the file name, and I was wondering if we should split them in separate locale directory as other collections.

src/sitemap.njk
```diff
@@ -0,0 +1,19 @@
+---
+permalink: /sitemap.xml
+eleventyExcludeFromCollections: true
+---
+<?xml version="1.0" encoding="utf-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+    {% for page in collections.all %}
+    {%- if page.url and page.data.locale == "en-CA" -%}
```
chosww, line 8: Page.data.locale is undefined in all cases, so sitemap isn't generated properly. Removing the second check on locale fixes the issue it seems.

chosww:
@greatislander This PR looks good to me, I approve the changes.

PR #1730, opened 2024-02-28
feat: apply proper themes to the cards used in homepage (resolves #1700)

### LINKED ISSUE #1700: Image of cards on the main page disappears in different themes
## Describe the bug

Using the website in different themes provided by UIO removes image of cards on the main page. From what I see, classes from UIO removes background-image. 

## To Reproduce

1. Go to the home page
2. Scroll down to where the cards are, and see images are properly displayed
3. Change the themes of the page using UIO
4. See the images on the cards are gone

## Expected behavior

Each card image should have different color set for different themes. 

## Screenshots

Screen Shot 2024-02-20 at 10 50 16 AM

chosww:

## Description

See #1700 for details.

## Steps to test

1. Go to the homepage and scroll down to where the cards are
2. Change the UIO themes
3. Verify the images on the cards are changing colours based on the UIO themes

## Related issues

Resolves #1700


greatislander (changes requested):
This is looking good, the scale is still not quite right though @chosww. I think maybe you are still working on that part?

src/assets/styles/components/_cards.scss
```diff
@@ -41,6 +41,7 @@
 
 		h2 {
 			font-weight: 600;
+			line-height: 130%;
```
greatislander, line 44: It's more concise use a decimal for line height, e.g. `1.3`.

src/assets/styles/components/_cards.scss
```diff
@@ -55,34 +56,167 @@
 
 		p {
 			font-weight: 400;
+			line-height: 130%;
```
greatislander, line 59: Same as above.

src/assets/styles/components/_cards.scss
```diff
@@ -55,34 +56,167 @@
...
+		}
+
+		.card__content {
+			block-size: 100%;
+			inline-size: 100%;
+			left: 0;
+			position: absolute;
+			top: 0;
```
greatislander, line 67: You can write `inset: 0;` instead of using `left` and `top`, see: https://developer.mozilla.org/en-US/docs/Web/CSS/inset

src/assets/styles/abstracts/_variables.scss
```diff
@@ -73,7 +73,7 @@ $uio-themes: "bw" #000 #fff,
 $breakpoints: (
 	xs: 280,
 	sm: 640,
-	md: 76,
```
greatislander, line 76: Good  catch!

src/assets/styles/abstracts/_breakpoints.scss
```diff
@@ -2,7 +2,7 @@
 	$min: breakpoint-value($name, $breakpoints);
 
 	@if $min {
-		@media (width >= rem($min)) {
+		@media (width > rem($min)) {
```
greatislander, line 5: Can you give me some context for this change?

greatislander (changes requested):

src/assets/styles/abstracts/_breakpoints.scss
chosww: For the desired breakpoints, we need to use both breakpoint-up and breakpoint-down on lg breakpoint; however, both breakpoint rules includes lg with >= and <=, so breakpoint at exact point was not captured by both breakpoint-up and breakpoint-down, so the layout was off at the exact point. I changed breakpoint-up to not include specified breakpoint, so that can be handled under breakpoint-down.

src/assets/styles/components/_cards.scss
chosww: Nice, that's cleaner 👍

src/assets/styles/components/_cards.scss
```diff
@@ -55,34 +56,166 @@
...
+	}
+}
+
+@include breakpoint-down(sm) {
+	.homepage-cards {
+		align-items: center;
+		display: flex;
+		flex-direction: column;
+	}
+}
```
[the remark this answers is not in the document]
chosww: Thanks @greatislander! I totally forgot about this fundamental of CSS 🤦‍♂️

greatislander (approved):

PR #2524, opened 2026-07-06
fix: update video size in RLDN

chosww:

## Description

There was an adjustment to the video size in the Regulating the Digital Domain page in this commit: https://github.com/inclusive-design/wecount.inclusivedesign.ca/commit/817e38f91863bfd4e980f3882271fd77661738f3, however, the change was not including videos in the French page because of how the container class is defined based on page name.

## Steps to test

1. Go to /rldn/
2. See the video size takes about 59% of the screen width, or is the same size as the videos in /rtdd/

### DIFF
--- a/src/assets/styles/layout/_layout.css
+++ b/src/assets/styles/layout/_layout.css
@@ -261,7 +261,8 @@ figcaption {
 			inline-size: 100%;
 		}
 
-		.regulating-the-digital-domain .embed--youtube {
+		.regulating-the-digital-domain .embed--youtube,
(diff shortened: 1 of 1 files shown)


cindyli (approved):
The use of the classes `.regulating-the-digital-domain` and `.reguler-le-domaine-numerique` is fragile. These classes are system generated based on the page title that can be modified through the CMS. Any change to the page title may result in updated class names then break this implementation.

Since this pattern has been in use for years, I will proceed with merging this PR as is. But we should consider a more robust solution in the future.

